PR #15, opened 2019-12-05
Update controller with edge developer config file with ziti-fabric pa…

plorenz:
…th and remove postgresql section. Fixes GH-14

### DIFF
--- a/etc/ctrl.with.edge.yml
+++ b/etc/ctrl.with.edge.yml
@@ -10,10 +10,10 @@ v: 3
 db:                     ${ZITI_DATA}db/ctrl.db
 
 identity:
-  cert:                 ${ZITI_SOURCE}fabric/etc/ca/intermediate/certs/ctrl-client.cert.pem
-  server_cert:          ${ZITI_SOURCE}fabric/etc/ca/intermediate/certs/ctrl-server.cert.pem
-  key:                  ${ZITI_SOURCE}fabric/etc/ca/intermediate/private/ctrl.key.pem
-  ca:                   ${ZITI_SOURCE}fabric/etc/ca/intermediate/certs/ca-chain.cert.pem
+  cert:                 ${ZITI_SOURCE}ziti-fabric/etc/ca/intermediate/certs/ctrl-client.cert.pem
+  server_cert:          ${ZITI_SOURCE}ziti-fabric/etc/ca/intermediate/certs/ctrl-server.cert.pem
+  key:                  ${ZITI_SOURCE}ziti-fabric/etc/ca/intermediate/private/ctrl.key.pem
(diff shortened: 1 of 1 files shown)


etc/ctrl.with.edge.yml
```diff
@@ -69,11 +69,3 @@ edge:
       # (optional, defaults to 5) The length of time that a Ziti Edge Router enrollment should remain valid. After
       # this duration, the enrollment will expire and not longer be usable.
       durationMinutes: 5
-  persistence:
```
dovholuknf: should this be left around for a few more cycles? seems like this could be left commented out for historical reasons/migration?

etc/ctrl.with.edge.yml
```diff
@@ -10,10 +10,10 @@ v: 3
...
 
 identity:
-  cert:                 ${ZITI_SOURCE}fabric/etc/ca/intermediate/certs/ctrl-client.cert.pem
-  server_cert:          ${ZITI_SOURCE}fabric/etc/ca/intermediate/certs/ctrl-server.cert.pem
-  key:                  ${ZITI_SOURCE}fabric/etc/ca/intermediate/private/ctrl.key.pem
-  ca:                   ${ZITI_SOURCE}fabric/etc/ca/intermediate/certs/ca-chain.cert.pem
+  cert:                 ${ZITI_SOURCE}ziti-fabric/etc/ca/intermediate/certs/ctrl-client.cert.pem
+  server_cert:          ${ZITI_SOURCE}ziti-fabric/etc/ca/intermediate/certs/ctrl-server.cert.pem
+  key:                  ${ZITI_SOURCE}ziti-fabric/etc/ca/intermediate/private/ctrl.key.pem
+  ca:                   ${ZITI_SOURCE}ziti-fabric/etc/ca/intermediate/certs/ca-chain.cert.pem
```
michaelquigley: This syntax would be much nicer looking if it were:

`${ZITI_SOURCE}/ziti-fabric/...`

plorenz: I agree. It's the way it is b/c if ZITI_SOURCE isn't specified, it evaluates to blank and we end up with a relative path ziti-fabric/etc if there's no slash, or an absolute path /ziti-fabric/etc if there is.

If having the absolute path as a fallback is OK, I can switch it. 
I considered adding logic to check for empty ZITI_SOURCE but there's nothing special about it. Other people might use different values in their config files, so it seemed a little strange to add something to the code for it.

Thoughts?

etc/ctrl.with.edge.yml
plorenz: I'll add it back, commented out for now

PR #31, opened 2020-01-08
Fixes netfoundry/ziti-fabric#30

andrewpmartinez:


controller/controller.go
```diff
@@ -107,6 +107,12 @@ func (c *Controller) Shutdown() {
 		}
 	}
 
+	if c.config.Db != nil {
+		if err := c.config.Db.Close(); err != nil {
+			pfxlog.Logger().WithError(err).Error("failed to close db")
+		}
+	}
+
 	c.network.Shutdown()
```
michaelquigley: Wouldn't we want this to run _after_ the `network` has a chance to shut down? It doesn't depend on the database right now, but it probably will soon.

andrewpmartinez: I can def move it to a `defer`. Didn't consider the shutdown needing to r/w config.

michaelquigley: You don't necessarily need to use `defer`, couldn't you just move it to happen after `c.network.Shutdown()`?

andrewpmartinez: I could the only functional difference is if shutdown() panics.

PR #32, opened 2020-01-17
Phase 13: Chapter 5

michaelquigley:
The UDP implementations provided by the fabric have been refactored to include:

## xgress_proxy_udp

A proxy implementation (the listener is wired to a specific service in the configuration), that accepts packets from plain UDP clients. Only provides a listener, as the dialler from `xgress_transport_udp` can be used to support UDP termination for `xgress_proxy_udp` clients.

## xgress_transport_udp

A UDP listener (with a session request payload as the first packet), and dialer implementation.

## xgress_udp

Shared UDP packet session infrastructure. Used by both `xgress_proxy_udp` and `xgress_transport_udp`.


xgress_udp/address.go
```diff
@@ -29,27 +28,10 @@ var supportedNetworks = map[string]string{
...
-
-func (pa *packetAddress) Address() string {
-	return pa.address
-}
-
-func parseAddress(s string) (*packetAddress, error) {
+func Parse(s string) (*PacketAddress, error) {
 	tokens := strings.Split(s, ":")
 	if len(tokens) < 2 {
-		return nil, errors.New("invalid format")
```
plorenz: Is there a difference between errors.New and fmt.Errorf if there are no msg parameters?

michaelquigley: Haven't measured it. But I tend to like the consistency of just always using `fmt.Errorf`.

Issue #38, opened 2020-01-30
Service Binding 2


plorenz:
Currently we try to derive binding based on address when creating services. Should we do that dynamically when being dialed? 
If not, we need to enforce service binding so that edge cannot bind a non-edge service.

michaelquigley:
I'm not sure I'm following you. The binding is the binding. The address is the address. For plain fabric services, the binding is resolved and passed the address when dialing the termination of the service. The address needs to be in a format understood by the xgress component named by the binding.

The fabric should not understand anything about edge services. This sounds like something that should be layered on top of the fabric within the edge itself?

plorenz:
This will be resolved by the work to extract terminators from services. See https://github.com/netfoundry/ziti-fabric/wiki/Pluggable-Service-Terminators

PR #41, opened 2020-02-04
initial support for passing crypto material

ekoby:
this is initial crack at supporting crypto key exchange

More changes coming but early feedback is very welcome


pb/ctrl_pb/ctrl.proto
```diff
@@ -39,6 +40,7 @@ message BindRequest {
     BindType bindType = 1;
     string token = 2;
     string serviceId = 3;
+    bytes pubkey = 4;
```
ekoby: should this be more generic like SessionRequest? or vice-a-versa?

michaelquigley: Paul and I were just discussing that we need to revisit the generalized "bind" concept in the fabric at some point in the next chunk of time (#38).

I think this is probably okay as `bytes` for now.

controller/network/network.go
```diff
@@ -265,7 +265,7 @@ func (network *Network) UnbindService(srcR *Router, token string, serviceId stri
 	return errors.New("invalid service")
 }
 
-func (network *Network) CreateSession(srcR *Router, clientId *identity.TokenId, serviceId string) (*session, error) {
+func (network *Network) CreateSession(srcR *Router, clientId *identity.TokenId, serviceId string, hints map[uint32][]byte) (*session, error) {
```
michaelquigley: Is there some way we can make this optional without having to pollute every usage of it with `nil` that isn't concerned with "hints"?

What is this used for?

pb/ctrl_pb/ctrl.proto
ekoby: yep, I suspected as much, especially when we start thinking about clustered hosting

controller/network/network.go
michaelquigley: Like... couldn't we use `identity.TokenId` for this?

ekoby: `hints` is generic way of passing info between two endpoints. I intend to use that for key exchange. I wish go supported default parameter values, but no. We could duplicate the method but there was only one other use of it.

michaelquigley: I don't think "hints" is the right name for this concept.

ekoby: any suggestions?

michaelquigley: To me, this feels like something that should be pushed into `identity` somewhere, rather than done as a generic `map` in the fabric. If we had something like `identity.PeerKey` or similar, and that was part of the `CreateSession` call, that would make sense.

So yes... not into the name "hints", but even moreso, not really into just sticking a vague `map` here.

ekoby: A couple of points:
- I think we need a generic mechanism to pass peer info to the endpoint -- peer Id/app/version could be useful. We can even use it to implement transparent networking. 
- second, the key here is not really a peer key, it is a session key. Identity can have multiple sessions each starting with a new ephemeral key. Pushing ephemeral key support into identity framework does not make sense to me.

michaelquigley: `peerData` or `data` would probably be better. I just really don't like the name `hints`.

Also, you could certainly create your session, and then exchange your peer data. It doesn't necessarily _need_ to be layered into the initial session creation.

michaelquigley: This also might be a bit of premature optimization... sure having the fabric piggyback this data in the initial session creation might save you a couple of messages on the wire (you could do this as part of your protocol, above the fabric)... but as soon as you need to roll a key, you're going to have to do the key negotiation above the fabric anyway.

ekoby: seems that we can minimize impact with netfoundry/ziti-foundation#34

andrewpmartinez:
Can we close this? It is 22 days old and I'm sure it has moved on.

ekoby:
> Can we close this? It is 22 days old and I'm sure it has moved on.

ready for review

controller/handler_ctrl/bind_service_request.go
```diff
@@ -52,7 +52,7 @@ func (h *hostRequestHandler) HandleReceive(msg *channel2.Message, ch channel2.Ch
 
 	var err error
 	if request.BindType == ctrl_pb.BindType_Bind {
-		err = h.network.BindService(h.r, request.Token, request.ServiceId)
+		err = h.network.BindService(h.r, request.Token, request.ServiceId, request.Pubkey)
```
michaelquigley: Some places we have `Pubkey`, other places the `Key` is capitalized.

Would it be possible to just make the fabric-level API take an `interface{}`, and call it something more open-ended like `userData`? (I know you'd called it `hints` the first time).

ekoby: certainly we can do that, but I assume that this code would have to be refactored soon anyway for multi/HA-bindings (@plorenz ). 
for now we'd need to know to extract key anyway or implement a way to store any kind of data

michaelquigley: I'm sure it will be affected by those changes, but I'm also sure the concept will remain in the new implementation. So, we should still endeavor to find the right abstraction now.

plorenz: The HA work will add strategy options/inputs. At a first pass, what I'm thinking is that peerData could be map[string][]byte. The controller would extract the values it knows about and ignore anything else. So it wouldn't store arbitrary data.** Maybe on the controller side we deal with map[string][]byte, but in the SDK we provide specific APIs to allowing passing in pub key and strategyInput? That way the controller APIs are more static, and the SDK apis are more user friendly.

**Except strategyOptions are pretty arbitrary, have to figure out how to model that. Right now I'm thinking interpret the []byte as string JSON and decode it.

pb/ctrl_pb/ctrl.proto
```diff
@@ -73,6 +75,7 @@ message Route {
         string binding = 1;
         string address = 2;
         string destination = 3;
+        map<uint32, bytes> peerData = 4;
```
michaelquigley: Why is this `peerData` in some places and `hostData` in others? Would it make sense to just call ti `peerData` across the board, and ship it?

ekoby: `hostData` is provided by binding identity (hosting endpoint), `peerData` is send by connecting entity (client). Keeping them different makes a little more clear what's going on

michaelquigley: But from the perspective of the fabric, it's all the same thing... it's data that's being shuffled between pairs of endpoints? So that distinction kind of isn't really evident at the fabric layer?

pb/ctrl_pb/ctrl.proto
```diff
@@ -39,6 +40,7 @@ message BindRequest {
     BindType bindType = 1;
     string token = 2;
     string serviceId = 3;
+    map<uint32, bytes> hostData = 4;
```
michaelquigley: Also, we need to make sure we update the `Decoder` component `pb/ctrl_pb/decoder.go`... otherwise all of this new data will be lost when tracing.

pb/ctrl_pb/ctrl.proto
ekoby: replaced

pb/ctrl_pb/ctrl.proto
ekoby: added tracing of BindRequest,BindResponse, and peerData ids

controller/network/network.go
```diff
@@ -243,12 +243,16 @@ func (network *Network) LinkChanged(l *Link) {
...
+func (network *Network) BindService(srcR *Router, token string, serviceId string, peerData map[uint32][]byte) error {
+	log := pfxlog.Logger()
+
 	// 1: Find Service
 	if svc, found := network.serviceController.get(serviceId); found {
 		svc.Egress = srcR.Id
 		svc.EndpointAddress = "hosted:" + token
+		svc.PeerData = peerData
 
+		log.Debugf("binding service[%s] to session[%s] with hostdata[%v]", serviceId, token, svc.PeerData)
```
michaelquigley: `peerData`

xgress/request.go
```diff
@@ -197,12 +208,13 @@ func CreateSession(ctrl CtrlChannel, peer Connection, request *Request, bindHand
 	return &Response{Success: true, SessionId: sessionInfo.SessionId.Token}
 }
 
-func BindService(ctrl CtrlChannel, token string, serviceId string) error {
+func BindService(ctrl CtrlChannel, token string, serviceId string, hostData map[uint32][]byte) error {
```
michaelquigley: `peerData`

PR #50, opened 2020-02-27
Separate bolt stores into db package. 

plorenz:
Make stronger distinction between controllers and stores. Brings model closer to the way the edge works.


controller/network/network.go
```diff
@@ -96,35 +96,35 @@ func (network *Network) GetDb() boltz.Db {
 	return network.db
 }
 
-func (network *Network) GetRouterStore() RouterStore {
-	return network.routerStore
+func (network *Network) GetFabricStores() *db.Stores {
```
michaelquigley: Couldn't we just call this `GetStores`?

plorenz: I called it GetStores but then it conflicted with some GetStores methods in the edge. Changing this was a one-liner vs a larger change in edge.

michaelquigley: I don't see how it conflicts with anything here, it's a receiver function on `Network`?

plorenz: Edge also has a Stores and GetStores defined on an interface for various things, as well as an interface which represents network so it can be mocked for tests, so if something wants to be able to contain both edge stores and fabric stores, the names conflict.

plorenz: I'm not fond of it either, so I'm going to spend some time looking at how to restructure things in the next pass to be a little cleaner.

michaelquigley: 👍

PR #51, opened 2020-03-03
Add support for multiple endpoints to services

plorenz:


controller/db/router_store.go
```diff
@@ -36,23 +36,23 @@ type Router struct {
 	Fingerprint string
 }
 
-func (router *Router) GetId() string {
-	return router.Id
+func (entity *Router) GetId() string {
```
michaelquigley: `entity`? FWIW, I've been using `self` lately for receivers, so that I don't have to think of a good name for them.

michaelquigley: I'm guessing `Router` is an "entity", and you're using `entity` across all of the receivers of this type.

controller/network/network.go
```diff
@@ -246,28 +260,43 @@ func (network *Network) LinkChanged(l *Link) {
...
-	if svc, found := network.serviceController.get(serviceId); found {
-		svc.Egress = srcR.Id
-		svc.EndpointAddress = "hosted:" + token
-		svc.PeerData = peerData
+	if _, found := network.serviceController.get(serviceId); found {
+		endpoint := &Endpoint{
+			Id:        token,
+			Service:   serviceId,
+			Router:    srcR,
+			Binding:   "edge",
```
michaelquigley: Do we want this defaulting to `edge` in the fabric code?

pb/mgmt_pb/decoder.go
```diff
@@ -300,7 +300,7 @@ func (d Decoder) Decode(msg *channel2.Message) ([]byte, bool) {
 }
```
michaelquigley: We need to make sure we update the `Decoder` for the mgmt channel, and for anything new traversing the data plane or the control plane.

controller/db/router_store.go
plorenz: Exactly. I've been trying to do this specifically in places where there's a lot of copy and paste code, like database entities.

controller/network/network.go
plorenz: I was thinking about this and I think we can remove the Bind/Unbind concept from the fabric, since it's really an edge concept, and replace it with just add/remove endpoint. Then the edge is responsible for all the semantics, including binding and address format. 

TBD is how routers can inject their keys for end to end encryption. Either way, though, bind isn't an appropriate mechanism for that.

michaelquigley: That sounds like a good avenue to explore. Adding/removing an endpoint to reflect SDK-hosted service presence makes a _lot_ more sense to me.

PR #53, opened 2020-03-12
Add support for multiple endpoints to services and sync model abstrac…

plorenz:
…tions with edge


controller/network/network.go
```diff
@@ -243,100 +208,84 @@ func (network *Network) LinkChanged(l *Link) {
...
-	// 1: Find Service
-	if svc, found := network.serviceController.get(serviceId); found {
-		// 2: Allocate Session Identifier
-		sessionIdHash, err := network.sequence.NextHash()
-		if err != nil {
-			return nil, err
-		}
-		sessionId := &identity.TokenId{Token: sessionIdHash}
+	// 3: select endpoint
+	endpoint := svc.Endpoints[0]
```
michaelquigley: Selecting an endpoint before a router is selected _will_ have impacts on smart routing in general. We could very easily implement a strategy that always ends up selecting a poorly-performing router.

Not sure if there's anything to do right this second, other than just be aware that we're potentially neutering smart routing in its current form. The router selection process is being preempted by endpoint selection.

michaelquigley: An alternative may be something along the lines of: smart routing looks at the endpoints for a service, and then ranks those in performance order. The strategy would take that list as an input, and could override it, or it could choose to allow it to suggest a choice.

plorenz: I think we've got several options, including passing implicit or explicit weights from smart routing, as well as allowing endpoint selection to be deferred to smart routing, possibly by flipping the paradigm and having endpoint selection return a set of acceptable endpoints and then letting smart routing take over.

michaelquigley: Indeed. I'm sure we can figure out something that will work.

PR #58, opened 2020-03-20
Add a router start method

plorenz:


router/router.go
```diff
@@ -76,7 +76,7 @@ func (router *Router) RegisterXctrl(x xctrl.Xctrl) error {
 	return nil
 }
 
-func (r *Router) Run() error {
+func (r *Router) Start() error {
```
michaelquigley: Why?

plorenz: Gives you an option of whether you want a method which returns or not. For the test, I want control to return to the test. So if you want it to block, call Run, if you want it to return, call Start.

michaelquigley: Ok...

PR #71, opened 2020-03-31
expose IsConnected(routerId) for edge use

andrewpmartinez:


controller/network/network.go
```diff
@@ -587,6 +587,10 @@ func (network *Network) GetServiceCache() Cache {
 	return network.Services
 }
```
plorenz: Shouldn't need this. RouterController is already in Handlers for access by edge functionality.

andrewpmartinez: Ah I see, not `GetRouterController` but `appEnv.Handlers.Router.Routers`

PR #84, opened 2020-04-27
Add Xt (eXtensible Terminator strategies) framework

plorenz:


controller/xt_default/impl.go
```diff
@@ -0,0 +1,47 @@
+package xt_default
+
+import (
+	"github.com/netfoundry/ziti-fabric/controller/xt"
+	"github.com/pkg/errors"
+)
+
+const (
+	Name = "single-endpoint"
```
michaelquigley, line 9: Should we use `default`? `single-endpoint` is kind of wordy.

controller/xt_hs_random/impl.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+func NewFactory() xt.Factory {
+	return factory{}
+}
+
+type factory struct{}
+
+func (f factory) GetStrategyName() string {
+	return "hs-random"
```
michaelquigley, line 15: Maybe we could simplify `hs-random` to `random` and `xt_hs_random` to just `xt_random`?

controller/xt_hs_smartrouting/impl.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+func NewFactory() xt.Factory {
+	return factory{}
+}
+
+type factory struct{}
+
+func (f factory) GetStrategyName() string {
+	return "hs-smartrouting"
```
michaelquigley, line 15: same: `hs-smartrouting` -> `smartrouting`?

controller/xt_hs_smartrouting/impl.go
```diff
@@ -0,0 +1,68 @@
+package xt_hs_weighted
+
+import (
+	"github.com/netfoundry/ziti-fabric/controller/xt"
+	"math"
+)
+
+func NewFactory() xt.Factory {
```
michaelquigley, line 8: Wonder if it might be worth writing a short descriptive comment for each of these implementations, describing what they do?

controller/xt_default/impl.go
plorenz: It has both "" and "default" as aliases, but I'm ok making "default" the name and "" an alias. We could also just use "single" as the name. Thoughts?

controller/xt_hs_random/impl.go
plorenz: :+1:

controller/xt_hs_smartrouting/impl.go
plorenz: Will do

controller/xt_default/impl.go
michaelquigley: Not sure I like the concept of "aliases". Wouldn't it be better to have a single, specific name for each strategy?

plorenz: It's there almost exclusively to support "" as an alias so we can have a strategy which handles existing services with no terminator strategy defined. Another approach would be write a migration which converts all services with a strategy of "" to "default" and also add code which assigns default to new services which don't define a strategy. Thoughts?

michaelquigley: I feel like it would be better to do the migration, rather than building the migration into the running system (which is what the alias does)?  ...especially if we're already having to migrate data. Why not just fix the Xt name, also?

plorenz: Removing aliases. 

I'm now also debating about whether we even want a strategy which only allows a single terminator. Even services which only have a single actual server running may want to bind via multiple routers to reduce single points of failure. I'm thinking about dropping xt_default entirely and making smartrouting the default. Does that line of reasoning seem reasonable?

michaelquigley: I mean... if "default" and `smartrouting` do the same thing in the presence of a single terminator, then there's no reason to have both? Dropping the "default" implementation makes sense to me. So, `smartrouting` would become the default?

michaelquigley: There's also nothing to limit the number of terminators that can be specified, even though "default" will only ever pick the first one... so, it would be pretty easy to get into a potentially weird state there. So, yes... it sounds like we should just make `smartrouting` be the default.

plorenz: :+1: will take that path

plorenz:
> Other than a couple of minor things, this looks great.

Review comments hopefully all addressed. Let me know how it looks.

PR #85, opened 2020-04-27
Integrate Xt at store level. Add migration code.

plorenz:


controller/db/migration.go
```diff
@@ -1,20 +1,28 @@
...
 	if step.CurrentVersion == 1 {
-		return 1
+		stores.extractTerminators(step)
+		return 2
+	}
+
+	if step.CurrentVersion == 2 {
+		return 2
 	}
 	step.SetError(errors.Errorf("Unsupported fabric datastore version: %v", step.CurrentVersion))
```
michaelquigley, line 27: Super minor nit... but I'm pretty sure Goland's linter complains about error messages that start with a capital letter.

controller/db/terminator_store_test.go
```diff
@@ -92,6 +119,8 @@ func (ctx *TestContext) createTestTerminators() *terminatorTestEntities {
 	e.terminator2.Id = uuid.New().String()
 	e.terminator2.Service = e.service.Id
 	e.terminator2.Router = e.router2.Id
+	e.terminator2.Binding = uuid.New().String()
+	e.terminator2.Address = uuid.New().String()
```
michaelquigley, line 123: Weren't we using shortids for these? It's just a `_test`, but...

plorenz: These aren't ids, I just need some random text and uuid is a fast way to get some. Even in the cases where it is used for non-terminator ids, it's used for generating invalid ids (as opposed to unset ids). I removed the use for terminator ids

PR #86, opened 2020-04-27
Sync terminators when routers connect so invalid terminators can be r…

plorenz:
…emoved


controller/network/terminator.go
```diff
@@ -93,49 +112,36 @@ func (ctrl *TerminatorController) Create(s *Terminator) (string, error) {
...
-	if e.Binding == "" {
-		if strings.HasPrefix(e.Address, "udp:") {
-			e.Binding = "udp"
+func (ctrl *TerminatorController) checkBinding(terminator *Terminator) {
+	if terminator.Binding == "" {
+		if strings.HasPrefix(terminator.Address, "udp:") {
+			terminator.Binding = "udp"
 		} else {
-			e.Binding = "transport"
+			terminator.Binding = "transport"
```
michaelquigley, line 120: Is this going to break Xgress for non-`udp` and non-`transport` bindings?

plorenz: It shouldn't. It only takes effect if a binding is not specified. The edge is now aware of bindings, so a binding will only use this logic if the user doesn't provide a binding on purpose (as opposed to previously when they couldn't provide a binding).

Longer term, now that the edge can provide this, we could consider taking it out and letting the tooling do this.

PR #87, opened 2020-04-29
Integrate Xt into smartrouting/circuit creation. Includes several Xt cleanups

plorenz:
* Remove Preferred precedence, as this use case is covered by having static costs on terminators
* Use cost everywhere in XT. If a specific strategy cares about weights, it can convert costs to weights locally
* Expand cost range to uint16 from uint8
* Better weighted implementation
* Create failure cost tracking type so we know when to stop crediting for successes
* Share cost/credit accounting between smartrouting and weighted strategies


michaelquigley:
"weights -> costs"?

controller/network/path.go
```diff
@@ -78,13 +79,13 @@ func (network *Network) shortestPath(srcR *Router, dstR *Router) ([]*Router, err
 	routerPath = append(routerPath, dstR)
 
 	if routerPath[0] != srcR {
-		return nil, errors.New("not routable (~srcR)")
+		return nil, 0, fmt.Errorf("can't route from %v -> %v. source unreachable", srcR.Id, dstR.Id)
```
michaelquigley, line 82: So wordy. U n lik (~srcR)? 🤣

controller/xt_smartrouting/impl.go
```diff
@@ -28,7 +30,8 @@ const (
 /**
 The smartrouting strategy relies purely on maninpulating costs and lets the smart routing algorithm pick the terminator.
 It increases costs by a small amount when a new session uses the terminator and drops it back down when the session
-finishes. It also increases the cost whenever a dial fails.
+finishes. It also increases the cost whenever a dial fails and decreases it whenever a dial succeeds. Dial successes
+will only reduce costs by the amount that failures have previously increased it.
```
michaelquigley, line 34: Corner case that we probably don't need to care about right this second, but a terminator with intermittent failures will likely look the same cost-wise as one without any.

pb/ctrl_pb/ctrl.proto
```diff
@@ -30,13 +30,19 @@ message SessionRequest {
...
+}
+
 message CreateTerminatorRequest {
-    string id = 1;
     string serviceId = 2;
     string binding = 3;
     string address = 4;
     map<uint32, bytes> peerData = 5;
     uint32 cost = 6;
+    TerminatorPrecedence precedence = 7;
```
michaelquigley, line 45: I know "creating a terminator" is primarily to support the edge... why does the edge care about `TerminatorPrecedence`?

pb/mgmt_pb/mgmt.proto
```diff
@@ -126,6 +140,14 @@ message GetTerminatorResponse {
     Terminator terminator = 1;
 }
 
+message SetTerminatorCostRequest {
+    string terminatorId = 1;
+    TerminatorPrecedence precedence = 2;
+    uint32 staticCost = 3;
+    uint32 dynamicCost = 4;
+    int32 updateMask = 5;
```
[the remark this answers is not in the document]
plorenz: I could have three bool fields instead of one mask. Mask just seemed a bit tidier. In the end, it's very unlikely that anyone will ever see this b/c the mask will get set by the CLI on the other side.  If you feel strongly about it, I can change it

controller/xt_smartrouting/impl.go
plorenz: Yeah, figuring out how to account for that kind of thing is going to be fun. We'll have to see what testing and production scenarios turn up.

pb/ctrl_pb/ctrl.proto
plorenz: In an HA scenario the primary will want to indicate that that it should be required, and the secondary will want to be default precedence.

michaelquigley: Because the edge needs to make that decision at Xgress dial time, right?

pb/mgmt_pb/mgmt.proto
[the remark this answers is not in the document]
michaelquigley: I don't feel _strongly_ about it, or I wouldn't have approved the PR. :-)

I think I generally prefer the more obvious solution, even if it's more verbose. Separate bools seems like a simpler thing to read about in code versus looking at bit masks. But it's pretty straightforward in this case, so it's not a big deal.

pb/ctrl_pb/ctrl.proto
plorenz: Yes, the SDK will make a Bind call to the xgress_edge component and will include the cost and precedence. xgress_edge will that turn that into a create terminator call and pass along the cost and precedence.

pb/mgmt_pb/mgmt.proto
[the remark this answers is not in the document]
plorenz: This code will also likely be replaced relatively soon by our consolidated REST based management, which is the other reason I'm inclined to leave it :)

pb/ctrl_pb/ctrl.proto
michaelquigley: Makes sense. 👍

controller/xt_smartrouting/impl.go
michaelquigley: It'll likely end up wanting some kind of short term memory. We'll figure it out.

pb/mgmt_pb/mgmt.proto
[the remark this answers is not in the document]
michaelquigley: Indeed. 💯

PR #88, opened 2020-05-01
Move xt strategy init

plorenz:


controller/controller.go
```diff
@@ -52,6 +52,11 @@ type Controller struct {
 }
 
 func NewController(cfg *Config) (*Controller, error) {
+	xt.GlobalRegistry().RegisterFactory(xt_smartrouting.NewFactory())
+	xt.GlobalRegistry().RegisterFactory(xt_ha.NewFactory())
+	xt.GlobalRegistry().RegisterFactory(xt_random.NewFactory())
+	xt.GlobalRegistry().RegisterFactory(xt_weighted.NewFactory())
+
```
michaelquigley, line 59: Maybe we could stick these in a `registerXts` or `registerTerminators`, similar to `registerComponents` in the same file?

PR #91, opened 2020-05-06
Implement update terminator for ctrl channel. Lots of xt fixes. Log l…

plorenz:
…evel tidying


controller/xt_common/cost_visitor.go
```diff
@@ -30,25 +30,31 @@ func (visitor *CostVisitor) VisitDialSucceeded(event xt.TerminatorEvent) {
 			if visitor.SessionCost > credit {
 				increase := visitor.SessionCost - credit
 				if cost < (math.MaxUint16 - increase) {
+					// pfxlog.Logger().Infof("%v: dial+ %v -> %v", event.GetTerminator().GetId(), cost, cost+increase)
```
michaelquigley, line 33: Why leave these commented out here?

plorenz: i would leave them at trace, but logrus doesn't have trace. I'm pretty sure I'll want them again, but I don't really want them flooding the logs at debug.

Want to have a conversation about logging sometime soon. I feel like we're headed for a some debugging trouble when we can't conditionally enable logging for just parts of the application. I could be looking at it in the wrong way though... which is why I want to have the conversation :)

PR #96, opened 2020-05-26
netfoundry/ziti-fabric -> openziti/fabric

michaelquigley:
NetFoundry -> OpenZiti


docs/local_development.md
```diff
@@ -33,21 +33,21 @@ When you've got your `GOPATH` ready, you'll want to clone the repositories. I li
...
-git clone https://github.com/netfoundry/ziti-cmd.git
-git clone https://github.com/netfoundry/ziti-fabric.git
+git clone https://github.com/openziti/ziti.git
+git clone https://github.com/openziti/fabric.git
 ```
 
 We're going to need to update the `go.mod` file in the root of `ziti-cmd`, using the `replace` directive to point the `ziti-cmd` build at our local `ziti-fabric` development tree.
 
 ```
-cd ~/repos/ziti-cmd
```
dovholuknf, line 43: nice catch wrt ziti-cmd here...

michaelquigley: Tried to blast through the docs real fast.

Issue #101, opened 2020-06-18
Add name to service and router


plorenz:
For consistency with edge, add names to service and router. Depends on short ids being added first. Default name to id if no name is provided.

PR #102, opened 2020-06-18
Add names to service and router. Add more store tests. Fixes GH-101

plorenz:


controller/db/router_store_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+	t.Run("test create invalid api routers", ctx.testCreateInvalidRouters)
+	t.Run("test create router", ctx.testCreateRouters)
+	t.Run("test load/query routers", ctx.testLoadQueryRouters)
+	t.Run("test update routers", ctx.testUpdateRouters)
+	t.Run("test delete routers", ctx.testDeleteRouters)
+}
+
+func (ctx *TestContext) testCreateInvalidRouters(_ *testing.T) {
+	defer ctx.cleanupAll()
```
andrewpmartinez, line 43: If these tests fail and you don't use the `T` passed in here they will exit with weird errors as `ctx.Req` will call fail now outside of its own context. It is undefined behavior.

plorenz: Fixed

PR #103, opened 2020-06-19
PayloadBufferController Leak

michaelquigley:
`PayloadBufferController.EndSession` was not properly cleaning up `PayloadBuffer` instances at session end, causing any `Payload` references inside the `PayloadBuffer` to just sit around (leak).

This may have started around the time we enabled support for single-router operation, because it looks like we're indexing `PayloadBuffer` instances by `address`+`sessionId` instead of just `sessionId` (necessary to support 2 `PayloadBuffer` instances for a single session in a single router. When the change was made, it looks like `EndSession` was left to try and remove `PayloadBuffer` instances by just `sessionId`, which wasn't removing anything.

Added an additional `cmap.ConcurrentMap` index called `sessions` to track `bindingId`s (`address`+`sessionId`) for each session, allowing `EndSession` to easily retrieve the list of `bindingId`s that need to be cleaned up.

Also adjusted `handler_xgress.BindHandler` to respect the `retransmission` option, and not create a `PayloadBuffer` for an Xgress if retransmission is disabled.

### DIFF
--- a/router/xgress/payload_buffer.go
+++ b/router/xgress/payload_buffer.go
@@ -69,10 +69,10 @@ func (controller *PayloadBufferController) BufferForSession(sessionId *identity.
 
 func (controller *PayloadBufferController) EndSession(sessionId *identity.TokenId) {
 	if v, found := controller.sessions.Get(sessionId.Token); found {
-		logrus.Infof("cleaning up for [s/%s]", sessionId.Token)
+		logrus.Debugf("cleaning up for [s/%s]", sessionId.Token)
 		bufferIds := v.([]string)
 		for _, bufferId := range bufferIds {
-			logrus.Infof("removing bufferId [%s]", bufferId)
+			logrus.Debugf("removing bufferId [%s]", bufferId)
 			controller.buffers.Remove(bufferId)
 		}
 		controller.sessions.Remove(sessionId.Token)
@@ -171,8 +171,8 @@ func (buffer *PayloadBuffer) Close() {
 
 func (buffer *PayloadBuffer) run() {
 	log := pfxlog.ContextLogger("s/" + buffer.sessionId.Token)
-	defer log.Infof("[%p] exited", buffer)
-	log.Infof("[%p] started", buffer)
+	defer log.Debugf("[%p] exited", buffer)
+	log.Debugf("[%p] started", buffer)
 
 	lastDebug := info.NowInMilliseconds()
 	for {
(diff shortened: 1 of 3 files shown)


router/xgress/payload_buffer.go
```diff
@@ -32,15 +33,29 @@ type PayloadBufferForwarder interface {
...
 	forwarder PayloadBufferForwarder
 }
 
 func NewPayloadBufferController(forwarder PayloadBufferForwarder) *PayloadBufferController {
-	return &PayloadBufferController{
+	pbc := &PayloadBufferController{
 		buffers:   cmap.New(),
+		sessions:  cmap.New(),
 		forwarder: forwarder,
 	}
```
plorenz, line 46: Do we want to either remove this or add a config flag to control it?

michaelquigley: i'm still cleaning this up. wait until i add reviewers. :-)

router/xgress/payload_buffer.go
```diff
@@ -49,15 +64,30 @@ func (controller *PayloadBufferController) BufferForSession(sessionId *identity.
 		return i.(*PayloadBuffer)
 	}
 
-	buffer := NewPayloadBuffer(sessionId, controller.forwarder)
+	buffer := NewPayloadBuffer(controller, sessionId, controller.forwarder)
 	controller.buffers.Set(bufferId, buffer)
+
```
plorenz, line 57: Maybe use cmap.Upsert here? There's a chance of a read/write conflict here. Not sure how likely it actually is, given it would have to be two different things with the same session, and sequencing would likely prevent that happening

plorenz:
Nice work!

One other way to fix this would be to have separate PayloadBufferController for the initiating and terminating sides. That would simplify the maps back to one, with just session as the key, but you'd have to make sure you used the right controller in the rx code, which would complicate things there.

Not sure if it's an overall win or not, but thought I'd put it up as something to consider.

michaelquigley:
> Nice work!
> 
> One other way to fix this would be to have separate PayloadBufferController for the initiating and terminating sides. That would simplify the maps back to one, with just session as the key, but you'd have to make sure you used the right controller in the rx code, which would complicate things there.
> 
> Not sure if it's an overall win or not, but thought I'd put it up as something to consider.

That sounds a lot more complicated, for not a lot of benefit in the short term. I suspect some of this code will end up getting a fresh look when I'm done with `transwarp`.

michaelquigley:
> Not sure if it's an overall win or not, but thought I'd put it up as something to consider.

I also suspect that when we integrate some of the `transwarp` idioms for memory management, a lot of this code will change.

PR #107, opened 2020-07-27
adds support for e2e via peer data for openziti/edge/issues#173

andrewpmartinez:
- add xt.PeerData type
- convert all peer data maps to xt.PeerData
- allow xgress dials to return session runtime peer data

### DIFF
--- a/controller/db/terminator_store.go
+++ b/controller/db/terminator_store.go
@@ -43,7 +43,7 @@ type Terminator struct {
 	Binding  string
 	Address  string
 	Cost     uint16
-	PeerData map[uint32][]byte
+	PeerData xt.PeerData
 }
 
 func (entity *Terminator) GetCost() uint16 {
@@ -66,7 +66,7 @@ func (entity *Terminator) GetAddress() string {
 	return entity.Address
 }
 
-func (entity *Terminator) GetPeerData() map[uint32][]byte {
+func (entity *Terminator) GetPeerData() xt.PeerData {
 	return entity.PeerData
 }
 
--- a/controller/handler_ctrl/session_request.go
+++ b/controller/handler_ctrl/session_request.go
@@ -54,9 +54,17 @@ func (h *sessionRequestHandler) HandleReceive(msg *channel2.Message, ch channel2
 			if session, err := h.network.CreateSession(h.r, id, request.ServiceId); err == nil {
 				responseMsg := ctrl_msg.NewSessionSuccessMsg(session.Id.Token, session.Circuit.IngressId)
 				responseMsg.ReplyTo(msg)
+
+				//static terminator peer data
 				for k, v := range session.Terminator.GetPeerData() {
 					responseMsg.Headers[int32(k)] = v
 				}
+
+				//runtime peer data
+				for k, v := range session.PeerData {
+					responseMsg.Headers[int32(k)] = v
+				}
+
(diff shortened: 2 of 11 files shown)


plorenz:
Overall, looks pretty good. Two questions: 

1. Should we have a specific return type for the Route message which includes the PeerData, rather than putting it in the headers? Or do we want to keep it generic for extensibility and reducing coupling?
2. Can you convert xgress_edge to use the same mechanism? I don't think it would be a big lift and I think they should have the same behavior (since they can). We could keep the peerdata pieces around for while, then drop then after some grace period.

andrewpmartinez:
> Overall, looks pretty good. Two questions:
> 
> 1. Should we have a specific return type for the Route message which includes the PeerData, rather than putting it in the headers? Or do we want to keep it generic for extensibility and reducing coupling?

I thought about that and I was interested to see if anyone said anything. Right now there isn't a clear line between headers and peerData at all points. I'd be fine with doing the work to completely separate peerData from headers. However, that would mean introducing peerData as a protobuff field for all messages or introducing specific message types to handle it. I think at this point the fabric is onboard with having peerData as a concept. A succinct followup question is: do we add it to all messages even if it isn't used on most of them?

> 2. Can you convert xgress_edge to use the same mechanism? I don't think it would be a big lift and I think they should have the same behavior (since they can). We could keep the peerdata pieces around for while, then drop then after some grace period.

Sure could.

michaelquigley:
> I thought about that and I was interested to see if anyone said anything. Right now there isn't a clear line between headers and peerData at all points. I'd be fine with doing the work to completely separate peerData from headers. However, that would mean introducing peerData as a protobuff field for all messages or introducing specific message types to handle it. I think at this point the fabric is onboard with having peerData as a concept.

I would definitely prefer that we not do that. "Peer data" is a bootstrapping concept just used at session startup. Anything that would need to go across the data or control plane otherwise should go through higher-level data or control plane messages.

> A succinct followup question is: do we add a peerData field to all messages even if it isn't used on most of them?

Let's please not.
 
And I think we just leave this as a `map` for now until we've got another need.

plorenz:
> > Overall, looks pretty good. Two questions:
> > 
> > 1. Should we have a specific return type for the Route message which includes the PeerData, rather than putting it in the headers? Or do we want to keep it generic for extensibility and reducing coupling?
> 
> I thought about that and I was interested to see if anyone said anything. Right now there isn't a clear line between headers and peerData at all points. I'd be fine with doing the work to completely separate peerData from headers. However, that would mean introducing peerData as a protobuff field for all messages or introducing specific message types to handle it. I think at this point the fabric is onboard with having peerData as a concept.
> 
> A succinct followup question is: do we add a peerData field to all messages even if it isn't used on most of them?
> 
> > 1. Can you convert xgress_edge to use the same mechanism? I don't think it would be a big lift and I think they should have the same behavior (since they can). We could keep the peerdata pieces around for while, then drop then after some grace period.
> 
> Sure could.

So I wasn't proposing adding PeerData to all messages. However, the Route message now explicitly checks for PeerData in the response. It feels like, since it's not completely transparent we should have an explicit RouteResponse which makes it clear from the message that it can contain PeerData. 

On the other hand, it's still an opaque field, we're just shuffling it from header to header, so maybe it doesn't to be called out in that way.

PR #120, opened 2020-08-24
adds session remove handler for openziti/edge#285

andrewpmartinez:


plorenz:
There's an existing events framework for fabric sessions in network/session_lifecycle.go. We should probably use that (or update it if it's missing some essential functionality)

andrewpmartinez:
> There's an existing events framework for fabric sessions in network/session_lifecycle.go. We should probably use that (or update it if it's missing some essential functionality)

Done. Needed to add `clientId` to the delete event, other than that, it is good.

Issue #123, opened 2020-08-31
Scale latency metric when used for path costing


plorenz:
Using nanoseconds causes latency to have an outsize affect on path selection

Issue #122, opened 2020-08-31
If link vanishes during reroute, controller can panic


plorenz:
```[34m[ 171.041][39m [37m   INFO[39m [36mfabric/controller/network.(*Network).LinkConnected[39m: link [l/ynWJ] failed
[34m[ 171.041][39m [37m   INFO[39m [36mfabric/controller/handler_ctrl.(*faultHandler).HandleReceive [ch{FsfkfA4GR}->u{classic}->i{80jk}][39m: link fault [l/ynWJ]
[34m[ 171.041][39m [37m   INFO[39m [36mfabric/controller/network.(*Network).Run[39m: changed link [l/ynWJ]
[34m[ 171.041][39m [37m   INFO[39m [36mfabric/controller/network.(*Network).rerouteLink[39m: link [l/ynWJ] changed
[34m[ 171.042][39m [37m   INFO[39m [36mfabric/controller/network.(*Network).LinkConnected[39m: link [l/3Nr3] failed
[34m[ 171.042][39m [37m   INFO[39m [36mfabric/controller/handler_ctrl.(*faultHandler).HandleReceive [ch{6dBmMM4MR}->u{classic}->i{8ZAr}][39m: link fault [l/3Nr3]
[34m[ 171.042][39m [37m   INFO[39m [36mfabric/controller/network.(*Network).Run[39m: changed link [l/3Nr3]
[34m[ 171.042][39m [37m   INFO[39m [36mfabric/controller/network.(*Network).rerouteLink[39m: link [l/3Nr3] changed
[34m[ 171.042][39m [37m   INFO[39m [36mfabric/controller/network.(*Network).rerouteLink[39m: session [s/KAz7] uses link [l/3Nr3]
[34m[ 171.042][39m [33mWARNING[39m [36mfabric/controller/network.(*Network).rerouteSession[39m: rerouting [s/KAz7]
[34m[ 171.042][39m [31m  ERROR[39m [36mfoundation/channel2.(*channelImpl).rxer [ch{NJFxCLVMR}->u{classic}->i{QJjQ}][39m: rx error (short read)
[34m[ 171.043][39m [37m   INFO[39m [36mfabric/controller/handler_ctrl.(*xctrlCloseHandler).HandleClose [ch{NJFxCLVMR}->u{classic}->i{QJjQ}][39m: closing Xctrl instances
[...]

Issue #121, opened 2020-08-27
Terminators and Fixed Link Cost, Incorrect Path Selection


michaelquigley:
Darius reported an issue where he has two links with two terminators:

```
Id     | Src                      -> Dst                      | State        | Cost | Latency
ARBA   | 09TjLwHMR                -> OoJQ-wHGR                | Connected    | 100  | 0.0433 0.0433 ->down<-
AOq7   | CZejYQNMg                -> 09TjLwHMR                | Connected    | 1    | 0.0632 0.0630 ->down<-
7Ww7   | CZejYQNMg                -> M7ZjLwNGR                | Connected    | 1    | 0.0740 0.0740 ->down<-
7DzA   | CZejYQNMg                -> OoJQ-wHGR                | Connected    | 1    | 0.0642 0.0642 ->down<-
nPQ7   | M7ZjLwNGR                -> 09TjLwHMR                | Connected    | 1000000 | 0.0430 0.0430
A0J7   | M7ZjLwNGR                -> CZejYQNMg                | Connected    | 1    | 0.0742 0.0742
nLxn   | M7ZjLwNGR                -> OoJQ-wHGR                | Connected    | 1    | 0.0230 0.0229 ->down<-
nQmn   | OoJQ-wHGR                -> 09TjLwHMR                | Connected    | 1    | 0.0433 0.0433 ->down<-
ldQA   | OoJQ-wHGR                -> CZejYQNMg                | Connected    | 1    | 0.0642 0.0641 ->down<-
nMQn   | OoJQ-wHGR                -> M7ZjLwNGR                | Connected    | 1    | 0.0239 0.0238 ->down<- (edited) 

Terminators: (2)
Id           | Service      | Binding      | Destination
KkV9         | S9XlUwNMg    | edge_transport | 09TjLwHMR    -> tcp:10.10.255.4:22
pw7p         | S9XlUwNMg    | edge_transport | CZejYQNMg    -> tcp:10.10.255.3:22

Services: (1)
Id           | Name         | Terminator Strategy | Destination(s)
S9XlUwNMg    | MicroEdgeSSH | smartrouting | 09TjLwHMR    -> tcp:10.10.255.4:22
             |              | CZejYQNMg    -> tcp:10.10.255.3:22
```

Despite setting the cost to `1000000` on the link `nPQ7` and having the `smartrouting` terminator selection strategy, he is still being directed to the higher cost terminator:

```
Sessions: (1)
Id           | Client       | Service      | Path
[...]

Issue #112, opened 2020-08-04
Persist terminator precedence


plorenz:
If we don't, we lose important context, especially for router terminated connections.

PR #125, opened 2020-08-31
Persist terminator precedence. Fix link cost use/scaling. Fix link re…

plorenz:
…moval related panic. Fix service cache clearing on terminator change.

Fixes GH-112
Fixes GH-121
Fixes GH-122
Fixes GH-123
Fixes GH-124


controller/network/network.go
```diff
@@ -526,18 +526,23 @@ func (network *Network) UpdateCircuit(circuit *Circuit) (*Circuit, error) {
...
+	}
 	return circuit2, nil
 }
 
-func (network *Network) setLinks(circuit *Circuit) {
+func (network *Network) setLinks(circuit *Circuit) error {
 	if len(circuit.Path) > 1 {
 		for i := 0; i < len(circuit.Path)-1; i++ {
 			if link, found := network.linkController.leastExpensiveLink(circuit.Path[i], circuit.Path[i+1]); found {
 				circuit.Links = append(circuit.Links, link)
```
plorenz, line 539: Return error instead of ignoring it when a link can't be found. If a link isn't found we will panic later as array indices won't line up.

controller/network/network.go
```diff
@@ -683,9 +688,9 @@ func (network *Network) AcceptMetrics(metrics *metrics_pb.MetricsMessage) {
 		metricId := "link." + link.Id.Token + ".latency"
 		if latency, ok := metrics.Histograms[metricId]; ok {
 			if link.Src.Id == router.Id {
```
plorenz, line 690: If we uses nanos in path calculation then the latency will overwhelm the other costs. Millis seems like a reasonable scale.

controller/network/path.go
```diff
@@ -50,6 +50,7 @@ func (network *Network) shortestPath(srcR *Router, dstR *Router) ([]*Router, int
 			if _, found := unvisited[r]; found {
 				cost := int64(r.CostFactor)
 				if l, found := network.linkController.leastExpensiveLink(r, u); found {
```
plorenz, line 52: We were including router cost factor (which can't actually be set anywhere) instead of link cost. Do we want to leave router cost factor here, and flesh it out later, or drop it?

controller/network/service.go
```diff
@@ -89,7 +89,18 @@ func (ctrl *ServiceController) newModelEntity() boltEntitySink {
 func (ctrl *ServiceController) terminatorChanged(params ...interface{}) {
 	for _, entity := range params {
 		if terminator, ok := entity.(*db.Terminator); ok {
```
plorenz, line 91: Terminator updates were being ignored sometimes because the patch didn't have the service so nothing was cleared from cache

controller/network/network.go
michaelquigley: Why not use fractional millis and keep the resolution, but have the magnitude you want?

controller/network/path.go
michaelquigley: I'm not sure what you mean by "router cost factor". That's not something that existed before `Xt`. Link cost has always been included and should continue to be included, as it's currently the main externally-configured cost input.

controller/network/network.go
michaelquigley: I'm betting that we're going to need to iterate on the scaling and combination of costs again. The current model seems a little tricky.

plorenz: I'll give the fractional millis a whirl. We may end up making the scaling factor configurable at some point, but I want to see how millis work for now.

controller/network/path.go
plorenz: Router has a CostFactor which has existed for as long back as I could trace it in git. It's not set by anything other than the constructor. I think we should probably drop it unless you think the implementation will be fleshed out soon.

michaelquigley: Agreed. Drop it.

controller/network/network.go
michaelquigley: Ok... But like I said, we're probably going to need to revisit all of this again at some point. I suspect I'll want to spend some time on this stuff when I'm back in the fabric for Smart Routing 2.

PR #128, opened 2020-09-17
Add support for addressable terminators

plorenz:


controller/db/service_store.go
```diff
@@ -17,33 +17,53 @@
...
 	"go.etcd.io/bbolt"
 )
 
 const (
-	EntityTypeServices             = "services"
-	FieldServiceTerminatorStrategy = "terminatorStrategy"
+	EntityTypeServices                     = "services"
+	FieldServiceTerminatorStrategy         = "terminatorStrategy"
+	FieldServiceIdentityAddressingAllowed  = "identityAddressingAllowed"
+	FieldServiceIdentityAddressingRequired = "identityAddressingRequired"
```
michaelquigley, line 33: This seems a little strange to me... allowed, required.

michaelquigley: Feels like it would be a little more in the spirit of the openness and extensibility of the fabric to just have an identity on the terminator, and if it's present, it can be used for addressing, and if it's not, then it can't. The allowed and required things feel a lot more like edge concepts to me.

plorenz: I see where you're coming from and we could certainly move these to the edge (albeit moderately painfully, since we don't have this information available at the point where it would need to be checked and I'll probably need to refactor the flow so we can do it an efficient way). 

I think these flags could certainly apply to fabric services. The motivations are as follows:
1. identity addressing allowed - for some services, identity addressing will be used but only in the future to group terminators for smarter routing. In those cases, we don't want to allow addressing by identity, since identity is being used by the controller, it's not for end-user use. On the other hand, the identities in these cases can also be made random, so that an end-user wouldn't be able to reasonably guess them. 

2. Identity addressing required - You could use endpoint addressable services on the fabric as well, such as an ssh service where each terminator goes to a different machine and the identity is the hostname. It wouldn't make sense to randomly connect to any terminator in the service, so it would be kind to the end-user to not allow this to prevent confusion. However, as you say, maybe this isn't true to the spirit of the fabric. 

Let me know if any of this has changed your mind, and I'll start investigating what it would take to move these checks to the edge.

michaelquigley: I don't think that really changes my mind. There isn't really any other "modal" behavior in the fabric. In general, changing behavior in the fabric is primarily done using strategies, not switches.

I think we want one implementation in the fabric that will use identity addressing all the time, and in cases where the it isn't being used, the identity components of the address are just left off. We probably should formalize some sort of scheme like `identity@service` or `service@identity`... and then `service` would continue to work as it always has.

controller/network/network.go
```diff
@@ -357,18 +359,28 @@ func (network *Network) CreateSession(srcR *Router, clientId *identity.TokenId,
 	}
 }
 
-func (network *Network) selectPath(srcR *Router, svc *Service) (xt.Strategy, xt.Terminator, []*Router, error) {
-	if len(svc.Terminators) == 0 {
-		return nil, nil, nil, errors.Errorf("service %v has no Terminators", svc.Id)
+func parseServiceAndIdentity(service string) (string, string) {
```
michaelquigley, line 362: Minor nitpick... but the function is named "service and identity" and returns "identity and service".

michaelquigley (approved):

controller/network/network.go
plorenz: Ah, good point. I'll fix that

PR #134, opened 2020-10-06
fixes #133

### LINKED ISSUE #133: Add version information to hello
Version information is becoming more necessary to deal with backward compatibility options between the SDK/Routers and Routers/Controller.  The channel2 framework can be expanded to support a version header that is implementation agnostic. This header can be used by the Fabric/Edge to signal version information during hello and hello ack.

andrewpmartinez:
- Adds version header to outgoing controller hellos and incomming router
  hellos

### DIFF
--- a/controller/controller.go
+++ b/controller/controller.go
@@ -33,6 +33,7 @@ import (
 	"github.com/openziti/fabric/controller/xt_weighted"
 	"github.com/openziti/fabric/controller/xtv"
 	"github.com/openziti/foundation/channel2"
+	"github.com/openziti/foundation/common"
 	"github.com/openziti/foundation/profiler"
 	"github.com/openziti/foundation/util/concurrenz"
 	"github.com/pkg/errors"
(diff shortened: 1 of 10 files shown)


controller/controller.go
```diff
@@ -53,7 +54,7 @@ type Controller struct {
 	isShutdown concurrenz.AtomicBoolean
 }
 
-func NewController(cfg *Config) (*Controller, error) {
+func NewController(cfg *Config, versionProvider common.VersionProvider) (*Controller, error) {
```
andrewpmartinez, line 57: VersionProvider allows us to invert who is providing the build version information (i.e. the `ziti` repo)

michaelquigley: Any reason why we wouldn't just have `NewController` just do `NewVersionProviderImpl` itself and call it a day? Not sure we _really_ need to pass that in?

andrewpmartinez: The provider has build values such as version, revision, etc. that are set externally. If we pulled from foundation none of those values would be available.

PR #139, opened 2020-10-16
do not drop packets with body of 0 length

ekoby:
this change allows forwarding of packets that only have headers and no body. It is needed in cases when an xgress component wants to implement signal/control messaging.


michaelquigley (approved):
Cross your fingers. _Seems_ reasonable to me. We're probably going to want to make sure we do as much testing as we can.

PR #143, opened 2020-12-08
switch self references to pointers so that fail count updates work

plorenz:


michaelquigley (approved):
Ah, that old chestnut. 👍 

(bonus points still available for changing `s` to `self` 😎)

controller/xt_ha/impl.go
```diff
@@ -40,33 +40,33 @@ func (f factory) GetStrategyName() string {
 }
 
 func (f factory) NewStrategy() xt.Strategy {
```
ekoby, line 42: do the same for `factory` methods?

plorenz: Not strictly necessary, since the factory carries no state, but I made it consistent along with the rest of the strategies.

PR #146, opened 2020-12-11
Set default for RetxStartMs and change name to match westworld 3 name

plorenz:


michaelquigley (approved):
Consistency. 👍 

(to be fair, all of the `dilithium`/`westworld` config/profile names are `snake_case` (`retx_start_ms`). But that would be inconsistent with the rest of the Xgress options. :-)

PR #153, opened 2021-01-18
Shutdown ack handler on link close. Remove unnecessary timers

plorenz:


michaelquigley (approved):
Test, and this is... 👍

PR #156, opened 2021-01-21
Configurable Termination/Route Creation Timeouts

michaelquigley:
Make route messages from the controller to the router, at route creation time, subject to configurable timeouts. There are separate timeouts for route creation for the terminator, and route creation for the intermediate hops.

### DIFF
--- a/controller/network/network.go
+++ b/controller/network/network.go
@@ -755,7 +755,7 @@ func sendRoute(r *Router, createMsg *ctrl_pb.Route, timeout time.Duration) (xt.P
 		return nil, fmt.Errorf("unexpected response type %v received in reply to route request", msg.ContentType)
 
 	case <-time.After(timeout):
-		pfxlog.Logger().Errorf("timed out waiting for response to route message from [r/%s] for [s/%s]", r.Id, createMsg.SessionId)
+		pfxlog.Logger().Errorf("timed out after %s waiting for response to route message from [r/%s] for [s/%s]", timeout, r.Id, createMsg.SessionId)
 		return nil, errors.New("timeout")
 	}
 }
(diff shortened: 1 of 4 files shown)


controller/network/network.go
```diff
@@ -754,7 +754,7 @@ func sendRoute(r *Router, createMsg *ctrl_pb.Route) (xt.PeerData, error) {
 		}
 		return nil, fmt.Errorf("unexpected response type %v received in reply to route request", msg.ContentType)
```
plorenz, line 756: Do we want to log what the timeout was on error, or just have people look it up in the controller config?

controller/network/options.go
```diff
@@ -27,11 +28,15 @@ type Options struct {
 		RerouteFraction float32
 		RerouteCap      uint32
 	}
+	TerminationTimeout time.Duration
+	RouteTimeout       time.Duration
 }
```
plorenz, line 34: I've been using time.ParseDuration(string), which allows specifying durations as 1s, 5m, etc. 
In this case, not sure it's worth it, since seconds is a very reasonable unit for the measure. Does let you use shorter names, though :)

plorenz (approved):

controller/network/network.go
michaelquigley: Implemented.

controller/network/options.go
michaelquigley: That makes sense. But I'd want to change `cycleSeconds` to match, and I don't want to introduce a breaking change to that right this second.

I feel like there's a pretty big refactoring coming up of the controller once we get to this next chunk of work. We should definitely hit it then.

PR #157, opened 2021-01-21
Xgress fixes

plorenz:


controller/network/smart.go
```diff
@@ -24,13 +24,17 @@ import (
...
 	/*
 	 * Order sessions in decreasing overall latency order
 	 */
 	sessions := network.GetAllSessions()
-	log.Debugf("observing [%d] sessions", len(sessions))
+	if len(sessions) > 0 {
+		log.Debugf("observing [%d] sessions", len(sessions))
+	} else {
+		log.Tracef("observing [%d] sessions", len(sessions))
+	}
```
michaelquigley, line 37: Why the conditional `Debugf`/`Tracef`?

michaelquigley (approved):
Seems reasonable. :+1:

controller/network/smart.go
plorenz: just to reduce the log noise when you're testing data-flow bits with debug on a system

PR #163, opened 2021-01-29
Add latency and message metrics to router control channels

plorenz:


controller/handler_ctrl/bind.go
```diff
@@ -33,33 +35,37 @@ func newBindHandler(router *network.Router, network *network.Network, xctrls []x
...
-	traceDispatchWrapper := trace.NewDispatchWrapper(bindHandler.network.GetEventDispatcher().Dispatch)
-	ch.SetLogicalName(bindHandler.router.Id)
-	ch.AddReceiveHandler(newSessionRequestHandler(bindHandler.router, bindHandler.network))
-	ch.AddReceiveHandler(newCreateTerminatorHandler(bindHandler.network, bindHandler.router))
-	ch.AddReceiveHandler(newRemoveTerminatorHandler(bindHandler.network))
-	ch.AddReceiveHandler(newUpdateTerminatorHandler(bindHandler.network))
-	ch.AddReceiveHandler(newLinkHandler(bindHandler.router, bindHandler.network))
-	ch.AddReceiveHandler(newFaultHandler(bindHandler.router, bindHandler.network))
-	ch.AddReceiveHandler(newMetricsHandler(bindHandler.network))
+func (self *bindHandler) BindChannel(ch channel2.Channel) error {
```
michaelquigley, line 38: Always warms my heart.

michaelquigley (approved):
Looks pretty reasonable so far.

As a next step, it might also be nice if the controller had latency metrics for the control channels of all of the routers, directly. That would measure how long it takes for a message to transit from the controller to the router over the control channel, and then receive attention at the `ReceiveHandler`.

plorenz:
> Looks pretty reasonable so far.
> 
> As a next step, it might also be nice if the controller had latency metrics for the control channels of all of the routers, directly. That would measure how long it takes for a message to transit from the controller to the router over the control channel, and then receive attention at the `ReceiveHandler`.

Looking at this again, I think I'm going to refactor this so that we have a latency response receiver and a single go-routine for sending all the latency messages. That way if the router doesn't support sending responses we won't be waiting around for something that never comes. 

By using ctrl channel specific logic we will also get direct access to the results and can hand them off to whatever is interested.

plorenz:
> > Looks pretty reasonable so far.
> > As a next step, it might also be nice if the controller had latency metrics for the control channels of all of the routers, directly. That would measure how long it takes for a message to transit from the controller to the router over the control channel, and then receive attention at the `ReceiveHandler`.
> 
> Looking at this again, I think I'm going to refactor this so that we have a latency response receiver and a single go-routine for sending all the latency messages. That way if the router doesn't support sending responses we won't be waiting around for something that never comes.
> 
> By using ctrl channel specific logic we will also get direct access to the results and can hand them off to whatever is interested.

Added some foundation logic which allows more async style latency send/receive and also a way to plug into generic latency handler, so we can use that directly if we so desire. See: https://github.com/openziti/foundation/pull/177

PR #165, opened 2021-02-02
More xgress close logic cleanup

plorenz:


michaelquigley (approved):
This is the "rx only" timeout situation we discussed the other day. Makes sense. :+1: 

One other minor thing we should consider changing while we're in here is the hard-coded 5 second timeout (line 126 in `forwarder.go`).

plorenz:
> This is the "rx only" timeout situation we discussed the other day. Makes sense. +1
> 
> One other minor thing we should consider changing while we're in here is the hard-coded 5 second timeout (line 126 in `forwarder.go`).

Done. Made that configurable.

PR #167, opened 2021-02-04
Eliminate memory leak (forwardTable concurrent-map)

rentallect:
Purge session from forwardTable during unrouteTimeout if no xgressForSession is found


plorenz (approved):

michaelquigley (approved):
👍

PR #170, opened 2021-02-05
Fix terminator patch to not update/clear peer data unless requested

plorenz:


dovholuknf (approved):
any doc needed to update too?

michaelquigley (approved):

plorenz:
> any doc needed to update too?

Doc update will be needed when we update the edge REST API

PR #171, opened 2021-02-05
Add validators

plorenz:


plorenz:
Thought of a better approach

PR #177, opened 2021-02-08
Parallel Route & Friends

michaelquigley:
Parallel route and related changes.


plorenz:
side note: I see you're using plain `logrus`. Is that new best practice, rather than using `pfxlog.Logger()`?

michaelquigley:
> side note: I see you're using plain `logrus`. Is that new best practice, rather than using `pfxlog.Logger()`?

That's what I've been doing, yes. `pfxlog` is pretty much just a log formatter these days. If you have a context to push, then using the `pfxlog` wrapper is handy.

Related: It may make sense to address some of the "content selection" (enabling/disabling log sections) as a `pfxlog` capability.

plorenz (approved):

controller/network/network.go
```diff
@@ -321,35 +341,44 @@ func (network *Network) CreateSession(srcR *Router, clientId *identity.TokenId,
...
+			allCleanups[k] = v
+		}
 		if err != nil {
-			strategy.NotifyEvent(xt.NewDialFailedEvent(terminator))
-			retryCount++
-			if retryCount > 3 {
-				return nil, err
-			} else {
+			logrus.Warnf("route attempt [#%d] for [s/%s] failed (%v)", attempt + 1, sessionId.Token, err)
+			attempt++
```
plorenz, line 358: redundant cast

michaelquigley: I don't see a cast on the lines you highlighted?

controller/network/network.go
```diff
@@ -321,35 +341,44 @@ func (network *Network) CreateSession(srcR *Router, clientId *identity.TokenId,
...
-			strategy.NotifyEvent(xt.NewDialSucceeded(terminator))
-		}
 
-		// 6: Create Intermediate Routes
-		for i := 0; i < len(circuit.Path)-1; i++ {
-			_, err = sendRoute(circuit.Path[i], rms[i], network.options.RouteTimeout)
-			if err != nil {
-				return nil, err
+			} else {
+				// revert successful routes
```
plorenz, line 363: not clear on why we cleanup if this is the last try, but not if it isn't

controller/network/network.go
michaelquigley: Ah, the line _below_.

controller/network/routesender.go
```diff
@@ -0,0 +1,168 @@
...
+
+	return peerData, nil, nil
+}
+
+func (self *routeSender) sendRoute(r *Router, routeMsg *ctrl_pb.Route) {
+	body, err := proto.Marshal(routeMsg)
+	if err != nil {
+		logrus.Errorf("error marshalling route message for [s/%s] to [r/%s] (%v)", routeMsg.SessionId, r.Id, err)
+		return
+	}
```
plorenz, line 144: Should we fast fail here if the control channel is closed and returns an error

controller/network/network.go
michaelquigley: Because the cleanup table is amended with the routers from each attempt. So the `allCleanups` map contains ALL the routers that would need to be cleaned up.

michaelquigley: And, cleaning up in intermediate attempts made for some strange races during testing.

controller/network/routesender.go
michaelquigley: Short answer: We can't, it's happning in a goroutine.

Long answer: We could, but this shouldn't happen, and if it does, it will get timed out. Implementing fail fast plumbing for this from a goroutine would add a lot of convolution for little gain, imo.

PR #179, opened 2021-02-09
Add default timeout for requests to the controller

plorenz:


router/xgress/options.go
```diff
@@ -46,6 +46,9 @@ type Options struct {
 	MaxCloseWait time.Duration
 
 	GetSessionTimeout time.Duration
+
+	// default timeout to use when sending requests to the controller
+	ControllerTimeout time.Duration
```
michaelquigley, line 51: Where is this used?

michaelquigley (approved):

router/xgress/options.go
plorenz: Refactoring xgress_edge to use it. xgress/request could be refactored to use it as well at some point.

michaelquigley: If it's in `openziti/fabric`, shouldn't we refactor something in `fabric` to use it?

plorenz: I may need to move it. The easier thing to refactor to use it doesn't have access xgress.Options. Maybe need to make it lop level router options, though not sure those are accessible in xgress implementations.

michaelquigley: Refactoring something like `xgress_proxy` to use `xgress.Options` doesn't seem like it would be a big deal?

PR #186, opened 2021-02-11
Forward Fault

michaelquigley:
Report forwarding faults to the controller for remediation.


router/forwarder/faulter.go
```diff
@@ -0,0 +1,90 @@
...
+		select {
+		case <-time.After(self.interval):
+			workload := self.sessionIds.Keys()
+			if len(workload) > 0 {
+				// Proactively remove from reported sessionIds. If we fail below, forwarder will continue to report.
+				for _, sessionId := range workload {
+					self.sessionIds.Remove(sessionId)
+				}
+
+				sessionIds := workload[0]
```
plorenz, line 67: `sessionIds := strings.join(workload, " ")`

router/xgress/retransmitter.go
```diff
@@ -142,10 +147,12 @@ func (retransmitter *Retransmitter) retransmitSender() {
 				if err := retransmitter.forwarder.ForwardPayload(retransmit.x.address, retransmit.payload); err != nil {
 					// if xgress is closed, don't log the error. We still want to try retransmitting in case we're re-sending end of session
 					if !retransmit.x.closed.Get() {
```
plorenz, line 149: Would it make sense to put the fault reporting in forward payload for simplicity? Or do we want it here since it will get hit eventually in the retransmit?

Need a way to let the LinkSendBuffer know as well, so it can back off the retransmits

michaelquigley: I don't think so. It looks like `ForwardPayload` is an interface, so we'd have to implement the fault reporting in everything that implements `ForwardPayload`?

I don't know that I'm super worried about backing off the retransmits for the first push of this. Once the mechanism is working, we can implement those kinds of tweaks, imo...

michaelquigley: Also, it's the retransmit loop that we need to break. So it seems like the most direct place to put it for now.

plorenz (approved):

PR #204, opened 2021-03-05
Idle Route Garbage Collection

michaelquigley:
Track route usage timestamps. Scan for idle routes. Send inquiries to the controller when idle routes are discovered.

Also includes configurable "dwell time" at Xgress dial, helping with debugging timing-related scenarios.

Controller messages:
```
[ 121.347]    INFO fabric/controller/handler_ctrl.(*sessionConfirmationHandler).HandleReceive: received session confirmation request from [r/001]
[ 121.347]    INFO fabric/controller/handler_ctrl.(*sessionConfirmationHandler).sendUnroute: sent unroute to [r/001] for [s/VQMV]
[ 123.247]    INFO fabric/controller/handler_ctrl.(*sessionConfirmationHandler).HandleReceive: received session confirmation request from [r/002]
[ 123.247]    INFO fabric/controller/handler_ctrl.(*sessionConfirmationHandler).sendUnroute: sent unroute to [r/002] for [s/VQMV]
[ 125.290]    INFO fabric/controller/handler_ctrl.(*sessionConfirmationHandler).HandleReceive: received session confirmation request from [r/003]
[ 125.291]    INFO fabric/controller/handler_ctrl.(*sessionConfirmationHandler).sendUnroute: sent unroute to [r/003] for [s/VQMV]
```

Router messages:
```
[ 120.005] WARNING fabric/router/forwarder.(*Scanner).scan: [s/VQMV] idle after [1m0s]
[ 120.005] WARNING fabric/router/forwarder.(*Scanner).scan: sent confirmation for [1] sessions
```


router/forwarder/scanner.go
```diff
@@ -0,0 +1,106 @@
...
+	for sessionId, ft := range sessions {
+		if time.Since(ft.(*forwardTable).last) > self.timeout {
+			idleSessionIds = append(idleSessionIds, sessionId)
+			logrus.Warnf("[s/%s] idle after [%s]", sessionId, self.timeout)
+		}
+	}
+
+	if len(idleSessionIds) > 0 {
+		logrus.Debugf("found [%d] idle sessions, confirming with controller", len(idleSessionIds))
+
```
plorenz, line 87: feels like self.ctrl should never be nil

router/forwarder/scanner.go
```diff
@@ -0,0 +1,106 @@
...
+		if self.ctrl != nil {
+			confirm := &ctrl_pb.SessionConfirmation{}
```
plorenz, line 89: Any reason not to just set `confirm.SessionIds = idleSessionIds`

router/forwarder/scanner.go
plorenz: Oh, I see, scanner is initialized then ctrl is set later.

Not for this PR, but it feels like it would be cleaner to init ctrl channel first,then init router

router/forwarder/tables.go
```diff
@@ -36,11 +37,13 @@ func newSessionTable() *sessionTable {
 }
 
 func (st *sessionTable) setForwardTable(sessionId string, ft *forwardTable) {
+	ft.last = time.Now()
 	st.sessions.Set(sessionId, ft)
 }
 
 func (st *sessionTable) getForwardTable(sessionId string) (*forwardTable, bool) {
 	if ft, found := st.sessions.Get(sessionId); found {
```
plorenz, line 45: This won't fix a garbage xgress with something stuck in retransmit, because the retransmitter will hit getForwardTable.

Could we key it off of the last time we got traffic from the xgress peer?

michaelquigley: It's not intended to fix that case.

Possibly. I'll look into it.

router/forwarder/scanner.go
michaelquigley: :+1:

router/forwarder/tables.go
michaelquigley: Keying off xgress doesn't fix middle-hop routers in any way.

plorenz (approved):

Issue #206, opened 2021-03-09
Parallelize Re-routing in network.Network


michaelquigley:
Blocking loops between routers/forwarding faults/re-routing in `network.Network.Run`. Parallelize re-routing activity, but control concurrency on sessions. The first re-routing activity wins, subsequent attempts, will exit.

plorenz:
Fixed in PR #207

PR #207, opened 2021-03-09
Parallel Network Re-route with Session Concurrency

michaelquigley:
See issue #206.

### DIFF
--- a/controller/network/network.go
+++ b/controller/network/network.go
@@ -737,7 +737,7 @@ func (network *Network) rerouteSession(s *Session) error {
 			return err
 		}
 	} else {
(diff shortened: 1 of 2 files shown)


controller/network/network.go
```diff
@@ -702,56 +708,68 @@ func (network *Network) rerouteLink(l *Link) error {
...
-		return nil
+			network.CircuitUpdated(s.Id, s.Circuit)
+
+			return nil
+		} else {
+			return err
+		}
 	} else {
-		return err
+		logrus.Warnf("not rerouting [s/%s], already in progress", s.Id.Token)
```
plorenz, line 740: Maybe make this an info?

michaelquigley: :+1:

PR #211, opened 2021-03-10
Add interval counts, per service, for dial success, timeout, fail, other

plorenz:
Can be configured as follows:

```
events:
  jsonLogger:
    subscriptions:
      - type: services
    handler:
      type: file
      format: json
      path: /tmp/ziti-events.log
```
Results in output like:

```
{
  "namespace": "service.events",
  "event_type": "service.dial.no_terminators",
  "service_id": "HSfgHzIom",
  "count": 8,
  "interval_start_utc": 1615400160,
  "interval_length": 60
}
{
  "namespace": "service.events",
  "event_type": "service.dial.success",
  "service_id": "HSfgHzIom",
  "count": 29,
  "interval_start_utc": 1615400160,
  "interval_length": 60
}
```

### DIFF
--- a/controller/network/routesender.go
+++ b/controller/network/routesender.go
@@ -17,6 +17,7 @@
 package network
 
 import (
+	"fmt"
 	"github.com/golang/protobuf/proto"
 	"github.com/openziti/fabric/controller/xt"
 	"github.com/openziti/fabric/pb/ctrl_pb"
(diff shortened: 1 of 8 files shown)


plorenz:
@mguthrie88 Should we add timestamp to these and usage events? They both have a version of timestamp in them, but this would let us know when the event was generated. Not sure how useful that is. Primary use case would be to see how long gap was between interval start in router and how long it took to get to the controller.

mguthrie88:
The timestamp you have should be fine.

mguthrie88 (approved):

controller/network/routesender.go
```diff
@@ -106,6 +110,7 @@ attendance:
 
 					if status.r == tr {
 						strategy.NotifyEvent(xt.NewDialFailedEvent(terminator))
```
michaelquigley, line 112: I feel a refactor coming on for a lot of these kinds of internals. This `routeSender.route`, `network.Run`, etc., should probably all have `interface{}` types that define extension points where we can insert these kinds of strategies (handlers).

I suspect we're probably going to end up wanting to clean up a bunch of these internals in this way. Eventually our `controller` and `router` stacks probably get built through some big IoC-style factory.

PR #220, opened 2021-03-18
If accept fails, ch will be nil

plorenz:


controller/handler_ctrl/accept.go
```diff
@@ -90,7 +90,9 @@ func (ctrlAccepter *CtrlAccepter) Run() {
 
 		} else {
 			log.Errorf("error accepting (%s)", err)
-			_ = ch.Close()
+			if err.Error() == "closed" {
```
michaelquigley, line 93: I get it. But I don't like it.

michaelquigley (approved):

controller/handler_ctrl/accept.go
plorenz: This is just reverting back to what was previously there. Longer term, we should probably implement a specific type we can test against.

PR #230, opened 2021-04-08
Allow way to notify router of rename

plorenz:
@michaelquigley This is likely incomplete. I started down the path of clearing the cache on update, but we compare routers directly in some places. I started trying to fix that, but that had large ripples. I settled on this as the smallest fix and though we should tackle any bigger cleanups as part of the distributed controller work. Thoughts?

### DIFF
--- a/controller/network/network.go
+++ b/controller/network/network.go
@@ -896,6 +896,20 @@ func (network *Network) GetServiceCache() Cache {
 	return network.Services
 }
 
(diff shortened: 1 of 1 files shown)


michaelquigley (approved):
Sounds like we probably _should_ fix it holistically as part of the "2.0" data model. I'm fine with the smallest possible fix for the current branch.

PR #231, opened 2021-04-08
add ability to clear the cached state of a router from bbolt

andrewpmartinez:
Will be used for enrollment extension where the fingerprint has been updated

### DIFF
--- a/controller/network/router.go
+++ b/controller/network/router.go
@@ -17,6 +17,7 @@
 package network
 
 import (
+	"github.com/michaelquigley/pfxlog"
 	"github.com/openziti/fabric/controller/db"
 	"github.com/openziti/fabric/controller/models"
 	"github.com/openziti/foundation/channel2"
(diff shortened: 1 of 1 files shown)


controller/network/router.go
```diff
@@ -168,3 +168,7 @@ func (ctrl *RouterController) readInTx(tx *bbolt.Tx, id string) (*Router, error)
 	ctrl.cache.Set(id, entity)
 	return entity, nil
 }
+
```
plorenz, line 171: This was the first thing I tried when fixing the name sync. Unfortunately it breaks lots of things b/c we cache Router instances in Link and in ConnectedCache. There are several places we compare Router instances directly and changing instances in this cache can cause routing failures. 

Two options to fix this are:
1. Remove Router instances from connected cache and links. This is complicated b/c we still need to cache transient data from the router and the control channel
2. Disconnect the router on router changes. I just thought of this as an option. The Router will reconnect and repopulate with correct. Will likely cause duplicate links to be created.

plorenz (changes requested):
See comment

controller/network/router.go
andrewpmartinez: Neither of those sounds great.

Is secret option 3 possible:

3. Update the fingerprint in the cached instance?

plorenz: That's what I did for the name. Probably the best option. Could extend the API I put in there for the name

andrewpmartinez: I pushed something that seems to work.

I considered adding a by-product of the router store, but I would have to pass a ref to the network down to the store instance or a delegate. I didn't know if it was worth it.

PR #232, opened 2021-04-22
Add events for router online/offline

plorenz:


michaelquigley:
Naively, this looks like it's just scaffolding for something else that generates router online/offline events? Does this function without the edge?

plorenz:
> Naively, this looks like it's just scaffolding for something else that generates router online/offline events? Does this function without the edge?

It's pure fabric. The events are generated by Network in fabric and the fabric events system can take those and dump them to a file or stdout in JSON or plain text. The events system is extensible so the edge can also inject events into it.

michaelquigley:
> It's pure fabric. The events are generated by Network in fabric and the fabric events system can take those and dump them to a file or stdout in JSON or plain text. The events system is extensible so the edge can also inject events into it.

If that's the case, then this PR is just adding representations of the data that didn't exist? I feel like the title might be kind of misleading...

plorenz:
> > It's pure fabric. The events are generated by Network in fabric and the fabric events system can take those and dump them to a file or stdout in JSON or plain text. The events system is extensible so the edge can also inject events into it.
> 
> If that's the case, then this PR is just adding representations of the data that didn't exist? I feel like the title might be kind of misleading...

The events are new to the events framework. The events are not new to Network. So however you want to describe that is ok.
I personally like to think that in the event of new events we like to eventually have an event where we add those events to the event subsystem :smiley:

michaelquigley (approved):
:+1:

mguthrie88 (approved):

PR #233, opened 2021-04-29
Remove eXtensible Terminator Validator framework (xtv) as it's now un…

plorenz:
…used.

terminator validation can be done in custom control channel handlers instead


michaelquigley (approved):
:+1:

PR #241, opened 2021-05-17
Add stackdump and clean shutdown on kill -3

plorenz:


plorenz:
Don't need this. You have to spend a special signal, so if you don't want stackdump, use other signal

PR #248, opened 2021-06-17
Add configurable router health check

plorenz:


michaelquigley:
Wouldn't it make more sense to have the controller ping the routers and then centralize all of the health checks in the controller?

plorenz:
> Wouldn't it make more sense to have the controller ping the routers and then centralize all of the health checks in the controller?

We had a case where the controller was locked up and no router events were generated. There might also be cases where the controller thinks things are fine, but the router is having trouble.

michaelquigley (approved):
:man_shrugging:

Issue #243, opened 2021-05-26
Latency Timeout Metrics


michaelquigley:
Capture latency probe success/failure (timeout) metrics.

plorenz:
When heartbeat responses don't come back within a timely fashion we now set the latency to a marker value of something like 88888888888. This both sets the latency high, so that the link is unlikely to be selected, unless it's the only choice, and acts as a sentinel value. Metric data can be check for the sentinel to build up a picture of when the link was unresponsive.

PR #249, opened 2021-06-19
Add controller health check

plorenz:


controller/config.go
```diff
@@ -257,5 +265,41 @@ func LoadConfig(path string) (*Config, error) {
 		}
 	}
```
plorenz, line 267: Experimenting with a more fluent config style, similar to the variables lib in fablab. Let me know what you think. It allows for a couple of different styes. The first is here to the next comment. This is already half the code that I initially had to configure these bits

controller/config.go
```diff
@@ -257,5 +265,41 @@ func LoadConfig(path string) (*Config, error) {
...
+			if val, found := boltCheckConfig.GetDuration("initialDelay"); found {
+				cfg.HealthChecks.BoltCheck.InitialDelay = val
+			}
+		}
+	}
+
+	if errHolder.HasError() {
+		return nil, err
+	}
+
```
plorenz, line 292: Here's the second style which is even more concise. Thoughts on if either, both or neither of these approaches look good?

health/handler.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func (self *HealthCheckApiHandler) RootPath() string {
+	return "/health-checks"
+}
+
+func (self *HealthCheckApiHandler) IsHandler(r *http.Request) bool {
+	return strings.HasPrefix(r.URL.Path, self.RootPath())
+}
+
```
plorenz, line 65: @andrewpmartinez tried to make this match the edge API, so output looks like:

```
{
    "data": {
        "checks": [
            {
                "healthy": true,
                "id": "bolt.read",
                "lastCheckDuration": "19.699µs",
                "lastCheckTime": "2021-06-18T21:48:37-04:00"
            }
        ],
        "healthy": true
    },
    "meta": {}
}
```
Let me know that works.

plorenz:
If this looks reasonable, I'll switch the router over to this style as well so that the router and controller are consistent. I moved xweb up a level in anticipation of that, so we can use it in the router as well.

controller/config.go
plorenz: Here's the draft PR for the config map code: https://github.com/openziti/foundation/pull/243
It'll probably need support for a few more types and additional refinement.

health/handler.go
andrewpmartinez: If this is a list, `data` should be the array. If this is a single object that is returned and there are other properties other than `checks` it is fine.

plorenz: There's a `healthy` which provides a quick path to the overall health at the same level as `checks`

andrewpmartinez: Just noticed your `lastCheckedTime` has a TZ. I think we try to output in UTC. Not a huge deal though.

router/router.go
```diff
@@ -430,24 +420,24 @@ const (
...
+func (self *Router) RegisterDefaultDebugOps() {
+	self.debugOperations[DumpForwarderTables] = self.debugOpWriteForwarderTables
+	self.debugOperations[UpdateRoute] = self.debugOpUpdateRouter
+	self.debugOperations[CloseControlChannel] = self.debugOpCloseControlChannel
+	self.debugOperations[OpenControlChannel] = self.debugOpOpenControlChannel
 }
 
-func (router *Router) RegisterDebugOp(opId byte, f func(c *bufio.ReadWriter) error) {
-	router.debugOperations[opId] = f
+func (self *Router) RegisterDebugOp(opId byte, f func(c *bufio.ReadWriter) error) {
```
andrewpmartinez, line 430: I don't mind `self`. Got used to it while working in C#. Goes against the language creator's style guide so I'm always torn on it.  No change needed, just musing. 

Did you disable the default warning that GoLand shows?

andrewpmartinez (approved):

router/router.go
plorenz: yes. @michaelquigley started using it and I got on board because it saves time trying to think of what to call everything. I read their rationale for why it was wrong, and thought it was pretty weak :)

PR #255, opened 2021-07-30
Don't allow multiple control channels active at same time

plorenz:


plorenz:
Fix duplicates existing functionality

PR #257, opened 2021-08-03
Rename circuit to path

plorenz:


plorenz:
Superceded by https://github.com/openziti/fabric/pull/258, which includes this commit

PR #268, opened 2021-09-10
Close send buffer when unrouted, since sends will no longer work

plorenz:
I tested this by adding a 100ms delay to the ack sender. I was able to see the log spam every time. This fixes it, and I believe should be safe. Since we've been unrouted (which we only do after the circuit is closed), sends will never work again, so we might as well shutdown the send buffer


michaelquigley (approved):
Seems reasonable? :shrug:

PR #273, opened 2021-10-04
added geneve option to router config

dariuszSki:
Related to the edge feature/geneve PR


plorenz:
This change needs to be in the ziti project, in `ziti-router/subcmd/run.go`

PR #277, opened 2021-10-14
Add support for plugins

plorenz:


andrewpmartinez (approved):
I'm luke warm on this feature due to the lack of Windows support for go plugins. That said, having it isn't a terrible option as long as developers know the limitations.

C-style plugins would be platform agnostic, but causes other problems (i.e. having to deal with c-go).

PR #284, opened 2021-11-12
Add initial AT infrastructure, largely copied from edge. Add mgmt cha…

plorenz:
…nnel test


tests/testdata/ca/intermediate/certs/ca-chain.cert.pem
```diff
@@ -0,0 +1,70 @@
+-----BEGIN CERTIFICATE-----
```
andrewpmartinez, line 1: It is possible to generate these in the tests rather than statically generating the files.

This is fine, but the certs will expire at some point.

andrewpmartinez (approved):

tests/testdata/ca/intermediate/certs/ca-chain.cert.pem
plorenz: Agree. We need to fix this for the edge as well.

PR #288, opened 2021-11-17
Add xweb based REST API for the fabric

plorenz:


controller/api_impl/base_entity_api_model.go
```diff
@@ -0,0 +1,122 @@
+/*
```
andrewpmartinez, line 1: Mmmmm again lots of code duped from the edge. Link builders and other base entity items should be consolidated over time.

controller/api_impl/router_router.go
```diff
@@ -0,0 +1,114 @@
...
+		return wrapper.WrapRequest(func(n *network.Network, rc api.RequestContext) { r.Create(n, rc, params) }, params.HTTPRequest, "", "")
+	})
+
+	//fabricApi.RouterUpdateRouterHandler = router.UpdateRouterHandlerFunc(func(params router.UpdateRouterParams) middleware.Responder {
+	//	return wrapper.WrapRequest(func(n *network.Network, rc api.RequestContext) { r.Update(n, rc, params) }, params.HTTPRequest, params.ID, "")
+	//})
+	//
+	//fabricApi.RouterPatchRouterHandler = router.PatchRouterHandlerFunc(func(params router.PatchRouterParams) middleware.Responder {
+	//	return wrapper.WrapRequest(func(n *network.Network, rc api.RequestContext) { r.Patch(n, rc, params) }, params.HTTPRequest, params.ID, "")
+	//})
```
andrewpmartinez, line 66: ?

specs/swagger.yml
```diff
@@ -0,0 +1,1397 @@
...
+paths:
+  ###################################################################
+  # Services
+  ##################################################################
+  '/services':
+    get:
+      summary: List services
+      description: |
+        Retrieves a list of service resources; supports filtering, sorting, and pagination. Requires admin access.
+      tags:
```
andrewpmartinez, line 30: What kind of auth is required? Just a client cert?

specs/swagger.yml
```diff
@@ -0,0 +1,1397 @@
...
+  ##################################################################
+  '/links':
+    get:
+      summary: List links
+      description: |
+        Retrieves a list of link resources; doesn not supports filtering, sorting, or pagination. Requires admin access.
+      tags:
+        - Link
+      operationId: listLinks
+      #      parameters:
```
andrewpmartinez, line 427: Documentation says it support filter/sort/pagination - but inputs are commented out

andrewpmartinez (approved):
See comments, minor changes

controller/api_impl/base_entity_api_model.go
plorenz: Agree. I did a first pass at reducing duplication, but there's lots left to do

controller/api_impl/router_router.go
plorenz: ziti-fabric doesn't allow updating routers. I've implemented it now.

specs/swagger.yml
plorenz: Yes, when in pure fabric mode only a valid client cert is required (like current ziti-fabric). When running with the edge it piggy backs on edge authentication and requires a valid api session. 

I was concerned about removing the security section, but from I can see in the edge code we just ignore that parameter and enforce the presence of a valid ztSession in our code, so I thought it was ok.

specs/swagger.yml
plorenz: Doc does it does NOT support those things (with a misspelling). Fix misspelling and removed comment out inputs

specs/swagger.yml
andrewpmartinez: Open API 2.0 / Swagger doesn't have a way to specify client cert auth. The only option you have is to document it on every endpoint with words.

andrewpmartinez: It is interesting that this API has two security modes though. I'm not sure how to document it w/o having two specs.

PR #289, opened 2021-11-19
Add routing perf benchmark and improve router perf

plorenz:
// original
BenchmarkShortestPathPerf-16    	       1	3773070830 ns/op
BenchmarkShortestPathPerf-16    	       1	3825162708 ns/op
BenchmarkShortestPathPerf-16    	       1	3657423204 ns/op
BenchmarkShortestPathPerf-16    	       1	3850618501 ns/op

// with pre-sized link slice
BenchmarkShortestPathPerf-16    	       1	3527057280 ns/op
BenchmarkShortestPathPerf-16    	       1	3532569248 ns/op
BenchmarkShortestPathPerf-16    	       1	3830881675 ns/op

// going straight to link slice
BenchmarkShortestPathPerf-16    	       1	2776857287 ns/op
BenchmarkShortestPathPerf-16    	       1	2769336745 ns/op
BenchmarkShortestPathPerf-16    	       1	2772600435 ns/op
BenchmarkShortestPathPerf-16    	       1	2918562399 ns/op

// using COW map of connected routers
no real change

// with router stored COW slices of related links
BenchmarkShortestPathPerf-16    	1000000000	         0.09486 ns/op
BenchmarkShortestPathPerf-16    	1000000000	         0.07627 ns/op
BenchmarkShortestPathPerf-16    	1000000000	         0.08958 ns/op
BenchmarkShortestPathPerf-16    	1000000000	         0.1065 ns/op

// With short-circuited Dijkstra 
BenchmarkShortestPathPerf-16    	1000000000	         0.05746 ns/op
BenchmarkShortestPathPerf-16    	1000000000	         0.05734 ns/op
BenchmarkShortestPathPerf-16    	1000000000	         0.05560 ns/op
BenchmarkShortestPathPerf-16    	1000000000	         0.05702 ns/op

### DIFF
--- a/controller/network/router.go
+++ b/controller/network/router.go
@@ -28,6 +28,8 @@ import (
 	"github.com/pkg/errors"
 	"go.etcd.io/bbolt"
 	"reflect"
+	"sync"
+	"sync/atomic"
 )
 
 type Router struct {
@@ -38,6 +40,7 @@ type Router struct {
 	Control            channel2.Channel
 	Connected          concurrenz.AtomicBoolean
 	VersionInfo        *common.VersionInfo
+	routerLinks        RouterLinks
 }
 
 func (entity *Router) fillFrom(_ Controller, _ *bbolt.Tx, boltEntity boltz.Entity) error {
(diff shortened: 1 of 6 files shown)


controller/network/route_perf_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"fmt"
+	"github.com/openziti/fabric/controller/db"
+	"github.com/openziti/foundation/identity/identity"
+	"github.com/openziti/foundation/transport/tcp"
+	"testing"
+)
+
+func BenchmarkShortestPathPerf(b *testing.B) {
```
dovholuknf, line 11: are there also correctness tests that are run? as someone not deeply familiar with these tests i'd feel better knowing those were there too :)

dovholuknf (approved):

controller/network/router.go
```diff
@@ -179,3 +239,45 @@ func (ctrl *RouterController) UpdateCachedFingerprint(id, fingerprint string) {
...
+	links := self.GetLinks()
+	newLinks := make([]*Link, 0, len(links)+1)
+	for _, l := range links {
+		newLinks = append(newLinks, l)
+	}
+	newLinks = append(newLinks, link)
+	self.Store(newLinks)
+}
+
+func (self *RouterLinks) Remove(link *Link) {
```
dovholuknf, line 268: should these functions have metrics - or are they somewhere that's just not obvious?

dovholuknf (approved):

controller/network/router.go
plorenz: We have some events for these -> router online/offline, which is probably the most common reason links are removed. Don't have actual link events. Should probably talk to ops and see how useful those would be

controller/network/route_perf_test.go
plorenz: Added correctness test to ensure new behavior matches old behavior

PR #290, opened 2021-11-19
Add routing perf test and fix link cost check

plorenz:


controller/network/debug.go
```diff
@@ -0,0 +1,78 @@
+package network
```
dovholuknf, line 1: needs header and a comment as to what this is all about. did you commit this intentionally? feels out of place - thus maybe a comment?

controller/network/path.go
```diff
@@ -234,5 +234,6 @@ func minCost(q map[*Router]bool, dist map[*Router]int64) *Router {
 			min = d
 		}
 	}
+	//Debugf("next router: %v, cost: %v, unvisited: %v\n", selected.Id, min, len(q))
```
dovholuknf, line 237: dead code should just be deleted?

controller/network/route_perf_test.go
```diff
@@ -0,0 +1,173 @@
+package network
```
dovholuknf, line 1: add copyright header pls

controller/network/route_perf_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+	var routerIdx int
+	addRouter := func() *Router {
+		router := newRouterForTest(fmt.Sprintf("router-%03d", routerIdx), "", transportAddr, nil)
+		network.Routers.markConnected(router)
+		routerIdx++
+		return router
+	}
+
+	for i := 0; i < 50; i++ {
```
dovholuknf, line 41: this feels like a function you'll use numerous times when testing. forgot to make that comment on the other pr. this seems like a util test function type stuff, no?

controller/network/route_perf_test.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	b.StartTimer()
+	srcRouter := routers[0]
+	replaceIdx := len(routers) - 1
+	for i := 1; i < len(routers); i++ {
+		p, c, err := network.shortestPath(srcRouter, routers[i])
+		ctx.NoError(err)
+
+		//fmt.Printf("{cost: %v, path:[]string{", c)
```
dovholuknf, line 145: note: dead code

controller/network/route_perf_test.go
```diff
@@ -0,0 +1,173 @@
...
+		{cost: 155, path: []string{"router-000", "router-056"}},
+		{cost: 280, path: []string{"router-000", "router-002", "router-055"}},
+		{cost: 198, path: []string{"router-000", "router-002", "router-054"}},
+		{cost: 308, path: []string{"router-000", "router-002", "router-053"}},
+		{cost: 273, path: []string{"router-000", "router-052"}},
+		{cost: 133, path: []string{"router-000", "router-051"}},
+		{cost: 141, path: []string{"router-000", "router-050"}},
+	}
+
+	b.StartTimer()
```
dovholuknf, line 138: just a comment: maybe StartTimer should "stop" the timer too? i keep seeing the 'StopTimer' function call at the top of these tests which I don't quite 'get why' other than just making sure that the setup bits aren't swept into the timer?

dovholuknf (approved):

controller/network/route_perf_test.go
plorenz: It's just how go benchmark stuff works. By default the whole function is timed, but you can stop the timer and restart when you're done with expensive initialization.

controller/network/path.go
plorenz: fixed

controller/network/route_perf_test.go
plorenz: fixed

dovholuknf (approved):

controller/network/route_perf_test.go
plorenz: extracted these out, added to a test helper, put in util_test with stuff from debug.go

controller/network/route_perf_test.go
plorenz: Fixed

controller/network/debug.go
plorenz: Renamed to util_test and added a comment. Added some other potentially common test bits there as well

PR #292, opened 2021-11-30
recieve handlers should never send as they may block forever if the u…

plorenz:
…nderlying channel closes

### DIFF
--- a/controller/handler_ctrl/create_terminator.go
+++ b/controller/handler_ctrl/create_terminator.go
@@ -44,12 +44,16 @@ func (h *createTerminatorHandler) ContentType() int32 {
 }
 
 func (h *createTerminatorHandler) HandleReceive(msg *channel2.Message, ch channel2.Channel) {
+	log := pfxlog.ContextLogger(ch.Label())
 	request := &ctrl_pb.CreateTerminatorRequest{}
 	if err := proto.Unmarshal(msg.Body, request); err != nil {
-		handler_common.SendFailure(msg, ch, err.Error())
+		log.WithError(err).Error("failed to unmarshal create terminator message")
 		return
 	}
+	go h.handleCreateTerminator(msg, ch, request)
+}
 
+func (h *createTerminatorHandler) handleCreateTerminator(msg *channel2.Message, ch channel2.Channel, request *ctrl_pb.CreateTerminatorRequest) {
 	if request.Cost > math.MaxUint16 {
 		handler_common.SendFailure(msg, ch, fmt.Sprintf("invalid cost %v. cost must be between 0 and %v inclusive", request.Cost, math.MaxUint16))
 		return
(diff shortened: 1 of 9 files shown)


controller/handler_ctrl/create_terminator.go
```diff
@@ -46,10 +47,13 @@ func (h *createTerminatorHandler) ContentType() int32 {
 func (h *createTerminatorHandler) HandleReceive(msg *channel2.Message, ch channel2.Channel) {
 	request := &ctrl_pb.CreateTerminatorRequest{}
 	if err := proto.Unmarshal(msg.Body, request); err != nil {
-		handler_common.SendFailure(msg, ch, err.Error())
+		log.WithError(err).Error("failed to unmarshal create terminator message")
```
andrewpmartinez, line 50: `SendResult` used to do the following:

```
	log := pfxlog.ContextLogger(ch.Label())
	if !success {
		log.Errorf("%v error (%s)", ch.LogicalName(), message)
	}
```

This new log message has less information in it. Was that intentional?

controller/handler_ctrl/inspect.go
```diff
@@ -42,10 +41,14 @@ func (h *inspectHandler) HandleReceive(request *channel2.Message, ch channel2.Ch
 
 	inspectRequest := &ctrl_pb.InspectRequest{}
 	if err := proto.Unmarshal(request.Body, inspectRequest); err != nil {
-		log.Errorf("unexpected error (%s)", err)
-		h.respondWithError(ch, request, fmt.Errorf("failed to decode request: (%v)", err).Error())
+		log.WithError(err).Error("failed to unmarshal inspect message")
```
andrewpmartinez, line 44: I see that handleInspect does reply but this if block simply returns and doesn't send a response.

So there is no one that was waiting for the response?

controller/handler_ctrl/create_terminator.go
```diff
@@ -24,6 +24,7 @@ import (
 	"github.com/openziti/fabric/controller/network"
 	"github.com/openziti/fabric/pb/ctrl_pb"
 	"github.com/openziti/foundation/channel2"
+	log "github.com/sirupsen/logrus"
```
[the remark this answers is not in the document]
plorenz: I had copied the log message from elsewhere. This code didn't already have a `log := pfxlog.ContextLogger...` so Goland 'helpfully' assumed I meant logrus with log. Thanks for catching that, it's now fixed.

controller/handler_ctrl/create_terminator.go
plorenz: Now that I've added the `log ;= pfxlog...` bit back in, there's more info than before. The context logger has the channel label which includes the channel logical name. We're also logging the error and some extra info about what kind of message we were expecting.

andrewpmartinez (approved):

PR #294, opened 2021-12-01
Fix fabric auth and add tests

plorenz:


tests/rest_auth_test.go
```diff
@@ -0,0 +1,39 @@
...
+func Test_ListServicesWithValidCert(t *testing.T) {
+	ctx := NewTestContext(t)
+	defer ctx.Teardown()
+	ctx.StartServer()
+	ctx.requireRestPort(5 * time.Second)
+	client := ctx.NewRestClientWithDefaults()
+	resp, err := client.R().Get("https://localhost:1281/fabric/v1/services")
+	ctx.Req.NoError(err)
+	ctx.Req.True(resp.IsSuccess())
+}
```
andrewpmartinez, line 39: test for if you send in no cert?

andrewpmartinez (approved):

tests/rest_auth_test.go
plorenz: Added test with no cert

PR #297, opened 2021-12-07
Ensure we don't return a nil control channel while starting up

plorenz:


router/router.go
```diff
@@ -90,6 +90,10 @@ func (self *Router) MetricsRegistry() metrics.UsageRegistry {
 }
 
 func (self *Router) Channel() channel2.Channel {
+	// if we're just starting up, we may be nil. wait till initialized
+	for self.ctrl == nil {
```
ekoby, line 94: should we have some MAX time after which we stop the process?

plorenz: If the control channel isn't established within a certain amount of time, the router will exit

ekoby: some other code is handling that?

ekoby (approved):

router/router.go
plorenz: will do. the initial control channel connect has a timeout

plorenz: Added comment

PR #300, opened 2021-12-16
Add more router connect error logging

plorenz:


mikegorman-nf (approved):

controller/handler_ctrl/connect.go
```diff
@@ -44,70 +44,72 @@ func NewConnectHandler(identity identity.Identity, network *network.Network, xct
...
+
+	if len(errorList) > 0 {
+		return errorList.ToError()
+	}
+
+	log := pfxlog.Logger().WithField("routerId", hello.IdToken)
+
 	fingerprint := ""
-	if certificates != nil {
+	if len(certificates) > 0 {
```
ekoby, line 79: redundant,  line 50

controller/handler_ctrl/connect.go
```diff
@@ -44,70 +44,72 @@ func NewConnectHandler(identity identity.Identity, network *network.Network, xct
...
+		Roots:         config.RootCAs,
+		Intermediates: x509.NewCertPool(),
+		KeyUsages:     []x509.ExtKeyUsage{x509.ExtKeyUsageAny},
+	}
+
+	var errorList errorz.MultipleErrors
+
+	for _, cert := range certificates {
+		if _, err := cert.Verify(opts); err == nil {
+			return nil
```
ekoby, line 66: this seem wrong

controller/handler_ctrl/connect.go
plorenz: fixed

controller/handler_ctrl/connect.go
plorenz: fixed, good catch

controller/handler_ctrl/connect.go
```diff
@@ -98,16 +59,54 @@ func (self *ConnectHandler) HandleConnection(hello *channel2.Hello, certificates
...
-	return errorList.ToError()
+	if !hasValidCert && len(errorList) > 0 {
+		return errorList.ToError()
+	}
+
+	log := pfxlog.Logger().WithField("routerId", hello.IdToken)
+
+	fingerprint := ""
+	log.Debugf("peer has [%d] certificates", len(certificates))
+	for i, c := range certificates {
```
ekoby, line 81: we should filter out certificates that failed the check (`cert.Verify()`)above. 
in general this logic is suspect if multiple certs are present, since we only take one fingerprint

ekoby (approved):

controller/handler_ctrl/connect.go
plorenz: I'm thinking we could check against all valid certs. That way a client could present old and new certs

plorenz: change made

controller/handler_ctrl/connect.go
```diff
@@ -98,16 +61,51 @@ func (self *ConnectHandler) HandleConnection(hello *channel2.Hello, certificates
...
+			log.Debugf("%d): peer common name [%s]", i, cert.Subject.CommonName)
 		} else {
 			errorList = append(errorList, err)
 		}
 	}
 
-	//goland:noinspection GoNilness
-	return errorList.ToError()
+	if len(errorList) > 0 {
+		return errorList.ToError()
```
ekoby, line 79: check for valid certs/fingerprints instead?
this will exit and ignore valid cert(s) if there is an invalid one.

plorenz: Fixed. Check is now if there are valid fingerprints

PR #302, opened 2022-01-06
sets content-length http response header

andrewpmartinez:
Setting the content-length header for HTTP responses allows the golang net/http library to avoid chunking the response.

### DIFF
--- a/controller/api/responder.go
+++ b/controller/api/responder.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"bytes"
+	"fmt"
 	"github.com/go-openapi/runtime"
 	"github.com/michaelquigley/pfxlog"
 	"github.com/openziti/fabric/controller/apierror"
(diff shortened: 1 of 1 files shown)


controller/api/responder.go
```diff
@@ -67,15 +69,26 @@ func (responder *ResponderImpl) Respond(data interface{}, httpStatus int) {
...
 func (responder *ResponderImpl) RespondWithProducer(producer runtime.Producer, data interface{}, httpStatus int) {
 	w := responder.rc.GetResponseWriter()
+	buff := &bytes.Buffer{}
+	err := producer.Produce(buff, data)
+
+	if err != nil {
+		pfxlog.Logger().WithError(err).
+			WithField("requestId", responder.rc.GetId()).
+			WithField("path", responder.rc.GetRequest().URL.Path).
+			Debug("could not respond, producer errored, possible missing response content")
```
dovholuknf, line 79: should this be at error for now?

plorenz (approved):

controller/api/responder.go
ekoby: if it fails to produce bytes we should set error on response, right?

andrewpmartinez: It gets weird in here. We are at a very low level and the probability that the producer will error is low.

How should I respond if the producer that producer the right content format fails to format? Should I call the producer again with the error record? Seems awkward if the producer is erroring.

I can send an error code and a plain/text body w/ the error message. Attempting to format the output w/ the same producer is dubious.

PR #303, opened 2022-01-06
adds an error log for HTTP timeouts

andrewpmartinez:


controller/api/timeouts.go
```diff
@@ -89,6 +89,12 @@ func (h *timeoutHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	case <-ctx.Done():
 		tw.mu.Lock()
 		defer tw.mu.Unlock()
+
+		pfxlog.Logger().WithFields(map[string]interface{}{
+			"url":    r.URL,
+			"method": r.Method,
+		}).Errorf("timeout for request hit, returning Service Unavailable 503")
```
dovholuknf, line 96: nit: it'd be better imo to use the constant but "meh".

PR #304, opened 2022-01-18
Update control channel to new channel API

plorenz:


controller/config.go
```diff
@@ -25,6 +25,7 @@ import (
 	"github.com/openziti/fabric/pb/ctrl_pb"
 	"github.com/openziti/fabric/pb/mgmt_pb"
 	"github.com/openziti/fabric/router/xgress"
+	"github.com/openziti/foundation/channel"
 	"github.com/openziti/foundation/channel2"
```
andrewpmartinez, line 29: I'm not clear on whether the channel2 import should still be here

plorenz: mgmt channel still uses channel2

PR #305, opened 2022-01-19
use idle timeout option for idle timeout for http servers in xweb

andrewpmartinez:


plorenz (approved):

xweb/server.go
```diff
@@ -121,7 +121,7 @@ func NewServer(webListener *WebListener, demuxFactory DemuxFactory, handlerFacto
 				Addr:         bindPoint.InterfaceAddress,
 				WriteTimeout: webListener.Options.WriteTimeout,
 				ReadTimeout:  webListener.Options.ReadTimeout,
-				IdleTimeout:  webListener.Options.WriteTimeout,
+				IdleTimeout:  webListener.Options.IdleTimeout,
```
andrewpmartinez, line 124: It won't solve world hunger but we are one step closer to perfection.

PR #306, opened 2022-01-20
Move some events/metrics code from foundation to fabric

plorenz:


event/events.go
```diff
@@ -0,0 +1,47 @@
+package event
```
plorenz, line 1: copied from foundation

event/events_test.go
```diff
@@ -0,0 +1,82 @@
+package event
+
```
plorenz, line 2: moved from foundation, no changes

metrics/channel_reporter.go
```diff
@@ -0,0 +1,54 @@
+/*
+	Copyright NetFoundry, Inc.
```
plorenz, line 2: moved from foundation, no changes

andrewpmartinez (approved):

PR #307, opened 2022-01-21
adds ability to update and save loaded configuration

andrewpmartinez:
- internal configuration values are stripped before save
- adds method to create a backup of the current configuration

### DIFF
--- a/router/config.go
+++ b/router/config.go
@@ -30,13 +30,33 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/pflag"
 	"gopkg.in/yaml.v2"
+	"io"
 	"io/ioutil"
+	"os"
 	"time"
 )
 
 const (
+	// FlagsCfgMapKey is used as a key in the source configuration map to pass flags from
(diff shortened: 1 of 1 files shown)


router/config.go
```diff
@@ -108,6 +131,112 @@ func (config *Config) SetFlags(flags map[string]*pflag.Flag) {
...
+// CreateBackup will attempt to use the current path value to create a backup of
+// the file on disk. The resulting file path is returned.
+func (config *Config) CreateBackup() (string, error) {
+	source, err := os.Open(config.path)
+	if err != nil {
+		return "", fmt.Errorf("could not open path %s: %v", config.path, err)
+	}
+	defer func() { _ = source.Close() }()
+
+	destPath := config.path + ".backup." + time.Now().Format("20060102150405")
```
dovholuknf, line 143: 20060102150405? Y U no const?

dovholuknf (approved):

router/config.go
andrewpmartinez: Done

PR #310, opened 2022-01-27
Add.controller.move.options

andrewpmartinez:
Adds two options:

1) in the ctrl listener, adds `newListener` to shadow the existing `listener` configuration value

```
# the endpoint that routers will connect to the controller over.
ctrl:
  listener: tls:127.0.0.1:6262
  options:
  # (optional) settings
    # set the maximum number of connect requests that are buffered and waiting to be acknowledged (1 to 5000, default 1000)
    #maxQueuedConnects:      50
    # the maximum number of connects that have  begun hello synchronization (1 to 1000, default 16)
    #maxOutstandingConnects: 100
    # the number of milliseconds to wait before a hello synchronization fails and closes the connection (30ms to 60000ms, default: 1000ms)
    #connectTimeoutMs:       3000
    # Sets the control channel write timeout. A write timeout will close the control channel, so the router will reconnect
    #writeTimeout: 15s
    newListener: tls:localhost:6262
```

2) adds `web` xweb bindpoint configuration to specify a `newAddress` to shadow the existing `address` value

```
    # bindPoints - required
    # One or more bind points are required. A bind point specifies an interface (interface:port string) that defines
    # where on the host machine the webListener will listen and the address (host:port) that should be used to
    # publicly address the webListener(i.e. mydomain.com, localhost, 127.0.0.1). This public address may be used for
    # incoming address resolution as well as used in responses in the API.
    bindPoints:
      #interface - required
      # A host:port string on which network interface to listen on. 0.0.0.0 will listen on all interfaces
      - interface: 127.0.0.1:1280

        # address - required
        # The public address that external incoming requests will be able to resolve. Used in request processing and
        # response content that requires full host:port/path addresses.
        address: 127.0.0.1:1280

        # newAddress - optional
        # A host:port string which will be sent out as an HTTP header "ziti-new-address" if specified. If the header
        # is present, clients should update location configuration to immediately use the new address for future
        # connections. The value of newAddress must be resolvable both via DNS and validate via certificates
        newAddress: localhost:1280
```

On router connect, the `newListener` value is sent to the connecting router in a new `Settings` message which allows for arbitrary settings to be sent to routers at any time.

On xweb request, a new header (defined as `ZitiCtrlAddressHeader = "ziti-ctrl-address"` ) is added to every response with the value of `https://<newAddress>`


plorenz:
Overall looks reasonable. Are we relying on external orchestration to handle using the updated control channel configuration?

andrewpmartinez:
> Overall looks reasonable. Are we relying on external orchestration to handle using the updated control channel configuration?

This alters both the configuration file and run time. To actually connect using it either a disconnect or a restart would have to happen.

dovholuknf:
Will you put the comments into one of the current controller config files commented out like we have been? mabye into `ctrl.with.edge.yml` ?

dovholuknf:
Are there any PKI-related requirements one should be aware of? such as the address that you are sending people to/redirecting should also be from the same trust anchor, right?

andrewpmartinez:
> Will you put the comments into one of the current controller config files commented out like we have been? mabye into `ctrl.with.edge.yml` ?

I sure can, just can't do it from this repo.

andrewpmartinez:
> Are there any PKI-related requirements one should be aware of? such as the address that you are sending people to/redirecting should also be from the same trust anchor, right?

There are - which will be documented in the ctrl.with.edge.yml and ctrl.yml example files.  They are:

1) The newAddress and newListener should be in the controller's cert as IP/DNS SANs
2) The newAddress should be externally routable to the controller instance

The first can be checked for - don't know if it should be a warning or a fatal. Leaning towards fatal. The other isn't reliably checkable.

andrewpmartinez:
> ...newListener should be in the controller's cert as IP/DNS SANs

Added validation for newLIstner values to check the server certificates SANs

plorenz (approved):

PR #313, opened 2022-01-31
Update links to use channel library

plorenz:


router/xlink_transport/bind.go
```diff
@@ -0,0 +1,13 @@
+package xlink_transport
```
dovholuknf, line 1: header comments

router/handler_link/bind.go
```diff
@@ -0,0 +1,127 @@
+package handler_link
```
dovholuknf, line 1: header comments

router/handler_link/bind.go
```diff
@@ -0,0 +1,127 @@
...
+	binding.AddErrorHandler(newErrorHandler(self.xlink, self.ctrl))
+	binding.AddTypedReceiveHandler(newPayloadHandler(self.xlink, self.forwarder))
+	binding.AddTypedReceiveHandler(newQueuingAckHandler(self.xlink, self.forwarder, closeNotify))
+	binding.AddTypedReceiveHandler(&latency.LatencyHandler{})
+	binding.AddTypedReceiveHandler(newControlHandler(self.xlink, self.forwarder, closeNotify))
+	binding.AddPeekHandler(metrics2.NewChannelPeekHandler(self.xlink.Id().Token, self.forwarder.MetricsRegistry()))
+	binding.AddPeekHandler(trace.NewChannelPeekHandler(self.xlink.Id().Token, ch, self.forwarder.TraceController(), trace.NewChannelSink(self.ctrl.Channel())))
+
+	if self.trackLatency {
+		latencyMetric := self.metricsRegistry.Histogram("link." + self.xlink.Id().Token + ".latency")
```
dovholuknf, line 76: is there any way to embed `self.xlink.Id().Token` as a tag or anything else? aren't unbounded names like this 'frowned on'?

router/handler_link/bind.go
```diff
@@ -0,0 +1,127 @@
...
+		go latency.ProbeLatencyConfigurable(config)
+	}
+
+	return nil
+}
+
+func (self *bindHandler) verifyLink(l xlink.Xlink, ch channel.Channel) error {
+	var fingerprints []string
+	for _, cert := range ch.Certificates() {
+		fingerprint := fmt.Sprintf("%x", sha1.Sum(cert.Raw))
```
dovholuknf, line 102: you'd think this was just some function. does it make sense to put somewhere that is reusable?

router/handler_link/bind.go
plorenz: Metric names have to be unique. This gets transformed when it goes through the events subsystem into a metric named `link.latency` with tags for the source and target routers.

dovholuknf (approved):
lgtm. few comments here/there

router/handler_link/bind.go
plorenz: I checked and there is an nfpem.FingerprintFromX509. I'll use that

router/xlink_transport/bind.go
plorenz: fixed

router/handler_link/bind.go
plorenz: fixed

router/handler_link/bind.go
plorenz: fixed

PR #316, opened 2022-02-09
Don't call message sends in control channel bind

plorenz:


controller/network/network.go
```diff
@@ -253,8 +253,9 @@ func (network *Network) ConnectRouter(r *Router) {
 	network.routerChanged <- r
 
 	for _, h := range network.routerPresenceHandlers {
-		h.RouterConnected(r)
+		go h.RouterConnected(r)
```
andrewpmartinez, line 256: This is an interesting little change. In other event callbacks, we have chosen not to put them in go routines and instead let the callback logic decide.

I am not advocating for change. I am highlighting it.

plorenz: My thinking was that because slowdowns here can delay accepting control channels, we should handle this asynchronously.

controller/network/network.go
```diff
@@ -253,8 +253,9 @@ func (network *Network) ConnectRouter(r *Router) {
...
 	}
+	go network.ValidateTerminators(r)
```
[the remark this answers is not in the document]
plorenz: I looked at doing it, but decided it didn't make sense. Other places we use presence handlers either because the logic is optional or because it's non-fabric code. I figured making this into a presence handler would just obscure things more.

PR #318, opened 2022-02-15
Allow routers more control over link management

plorenz:


controller/network/link_controller.go
```diff
@@ -46,6 +48,22 @@ func (linkController *linkController) has(link *Link) bool {
 	return linkController.linkTable.has(link)
 }
 
+func (linkController *linkController) routerReportedLink(linkId, linkType string, src, dst *Router) (*Link, bool) {
+	linkController.lock.Lock()
+	linkController.lock.Unlock()
```
camotts, line 53: Is locking and immediately locking here intentional? Just looked odd to me without a defer

plorenz: Good catch, thank you! Fixed.

camotts (approved):
Looks good and makes sense to me

PR #324, opened 2022-02-21
Added cost to routers and relevant persistence and path checking.

camotts:
Added a cost to routers which is added to the pathing cost calculation.


router/config.go
```diff
@@ -120,6 +122,7 @@ type Config struct {
 		}
 	}
 	Plugins []string
```
plorenz, line 124: Since cost is adjustable via the model we don't need it adjustable via the config file

camotts: removed!

controller/db/router_store.go
```diff
@@ -19,39 +19,44 @@ package db
...
 func (entity *Router) SetValues(ctx *boltz.PersistContext) {
 	entity.SetBaseValues(ctx)
 	ctx.SetString(FieldName, entity.Name)
 	ctx.SetStringP(FieldRouterFingerprint, entity.Fingerprint)
+	ctx.SetInt32(FieldRouterCost, int32(entity.Cost))
 }
 
 func (entity *Router) GetEntityType() string {
 	return EntityTypeRouters
 }
```
plorenz, line 55: Is this used anywhere? If yes, can we rename to GetCost(), otherwise, can we drop it?

andrewpmartinez (approved):

andrewpmartinez (changes requested):
See Pauls comment. I was approving tabs I had already looked at and mistook this for one of Paul's PRs.

specs/swagger.yml
```diff
@@ -1131,6 +1141,11 @@ definitions:
       fingerprint:
         type: string
         x-nullable: true
```
plorenz, line 1143: I think we should add cost to the required fields for router updates. I think fingerprint should probably be there as well. That's it's not there now it probably an oversight. In general we have most fields required for update, since if you only want to update some things, you use patch. Only concern there is backwards compatibility, but that shouldn't be a big concern here, since this API isn't widely used.

@andrewpmartinez Thoughts?

controller/db/router_store.go
camotts: Completely forgot to rename this, was added back when I was first learning the ins and outs for consistency. Its currently not used so will drop it.

specs/swagger.yml
camotts: Will add cost to be required. Did you want me to go ahead and add fingerprint there as well?

controller/network/network.go
```diff
@@ -19,6 +19,11 @@ package network
 import (
```
plorenz, line 19: reshuffled imports are ok, as long as they make some kind of sense, if import organization changes are the only changes in the file, please revert, just to minimize churn :)

specs/swagger.yml
plorenz: Yes, please

controller/network/network.go
camotts: Can do! Wasnt sure so left it in anway

plorenz (approved):
Looks good!

andrewpmartinez (approved):

PR #326, opened 2022-02-28
added allow traversal flag

camotts:


controller/network/path_test.go
```diff
@@ -302,3 +303,79 @@ func TestShortestPath(t *testing.T) {
 		9 + 20 + 21 + 4 // link2 cost and src and dest latency plus dest router cost
 	req.Equal(int64(expected), cost)
 }
```
plorenz, line 305: Can you add a test for if the non-traversable router is the only router connecting the initiating and terminating routers and make sure that circuit creation fails in that case? You'll likely have to change the path selection algorithm

camotts: I can! One clarification though. I didn't have it completely not connect as to not lock connections. Was that not something I need to worry about?

camotts: Turns out I had unintentionally had it work that way anyway. Completely missed the error throwing on costs over maxint32. Added a unit test to cover this case.

plorenz: :+1: 

Can you add a couple more test cases?

Both of these should work, as they're on the ends, not in the middle
Untraversable <-> Untraversable 
Untraversable <-> Traversable <-> Untraversable

camotts: Added

controller/network/path_test.go
```diff
@@ -302,3 +303,208 @@ func TestShortestPath(t *testing.T) {
...
+
+	link := newLink("l0")
+	link.SetStaticCost(2)
+	link.SetDstLatency(10 * 1_000_000)
+	link.SetSrcLatency(11 * 1_000_000)
+	link.Src = r0
+	link.Dst = r1
+	link.addState(newLinkState(Connected))
+	network.linkController.add(link)
+
```
plorenz, line 413: This test should pass. Even though r1 is not traversable, it's ok because the path is not traversing r1, it's terminating at r1.

controller/network/path_test.go
```diff
@@ -302,3 +303,208 @@ func TestShortestPath(t *testing.T) {
...
```
plorenz, line 452: Should also allow this path. r0 and r1 are initiating and terminating path, they're not being traversed

controller/network/path_test.go
```diff
@@ -302,3 +303,208 @@ func TestShortestPath(t *testing.T) {
...
+	link = newLink("l2")
+	link.SetStaticCost(2)
+	link.SetDstLatency(10 * 1_000_000)
+	link.SetSrcLatency(11 * 1_000_000)
+	link.Src = r1
+	link.Dst = r2
+	link.addState(newLinkState(Connected))
+	network.linkController.add(link)
+
```
plorenz, line 503: Should be allowed as well

controller/network/path_test.go
```diff
@@ -302,3 +303,208 @@ func TestShortestPath(t *testing.T) {
...
+	closeNotify := make(chan struct{})
+	defer close(closeNotify)
+
+	network, err := NewNetwork("test", nil, ctx.GetDb(), nil, NewVersionProviderTest(), closeNotify)
+	req.NoError(err)
+
+	addr := "tcp:0.0.0.0:0"
+	transportAddr, err := tcp.AddressParser{}.Parse(addr)
+	req.NoError(err)
+
```
plorenz, line 476: Can we get one more test just like this one, but with the middle router being non-traversable? That scenario should be non-routable.

controller/network/path_test.go
camotts: Ah I must have misunderstood the intent here. The way it was in my head is that data shouldn't traverse through that router. So we only care if the middle routes are nontraversable then?

controller/api_impl/router_api_model.go
```diff
@@ -55,9 +55,10 @@ func MapCreateRouterToModel(router *rest_model.RouterCreate) *network.Router {
 			Id:   stringz.OrEmpty(router.ID),
 			Tags: TagsOrDefault(router.Tags),
 		},
-		Name:        stringz.OrEmpty(router.Name),
-		Fingerprint: router.Fingerprint,
-		Cost:        uint16(Int64OrDefault(router.Cost)),
+		Name:           stringz.OrEmpty(router.Name),
+		Fingerprint:    router.Fingerprint,
+		Cost:           uint16(Int64OrDefault(router.Cost)),
```
plorenz, line 60: Missed this on an earlier pass. I think AllowTraversal should default to true (or reverse the flag to NoTraversal). Otherwise we breaks backwards compatibility. This will likely affect the edge changes as well.

PR #327, opened 2022-02-28
Added global constants files for use in this and other projects

gberl002:


controller/constants.go
```diff
@@ -0,0 +1,10 @@
+package controller
```
plorenz, line 1: Please move these to controller/config.go

controller/network/constants.go
```diff
@@ -0,0 +1,13 @@
+package network
```
plorenz, line 1: Please move to controller/network/options.go

router/forwarder/constants.go
```diff
@@ -0,0 +1,25 @@
+package forwarder
```
plorenz, line 1: Please move to router/forwarder/options.go

xweb/constants.go
```diff
@@ -0,0 +1,31 @@
+package xweb
```
plorenz, line 1: Please move to xweb/options.go

plorenz (changes requested):
Overall is fine, but please move constants to point of use, since each set is only used in a single file let's keep those consolidated.

plorenz (approved):

PR #329, opened 2022-03-09
Fix panic caused by race condition on startup

plorenz:


router/link_registry.go
```diff
@@ -29,6 +29,16 @@ type linkRegistryImpl struct {
 	ctrlCh channel.Channel
 }
 
+func (self *linkRegistryImpl) ControlChannel() channel.Channel {
+	// we may get link requests before the control channel is fully
+	// established. wait until it's set before we return. Will only
+	// happen right at startup
+	for self.ctrlCh == nil {
+		time.Sleep(30 * time.Millisecond)
```
dovholuknf, line 37: Shouldn't this be in a loop with a timeout? What is the machine is slow or bogged down?

dovholuknf (approved):

PR #334, opened 2022-03-17
Fix flaky test

plorenz:


tests/testutil/message_collector.go
```diff
@@ -25,6 +25,10 @@ type MessageCollector struct {
 }
 
 func (self *MessageCollector) HandleReceive(m *channel.Message, ch channel.Channel) {
+	if m.ContentType == -33 {
+		logrus.Debug("ignoring reconnect ping")
```
dovholuknf, line 29: is there something 'interesting' here at all? should an issue be filed to 'figure out why this happens'? feels... squishy to call this a 'fix' but - maybe i'm just not in the know and this makes perfect sense? so i'll play my role... and just question it :)

plorenz: The test expects messages to come back in a certain order. The reconnect ping message is unpredictable and isn't something we care about in the test. The test fails when we get a reconnect ping instead of the expected initial message after connection. Since we don't care about it, this now ignores the message.

PR #335, opened 2022-03-18
Route error codes

camotts:
Added handling for an error code header and sending tunnel local address information back up to the controller.

Edge routers can now send error messages through the `ctrl_msg.RouteResultErrorHeader`. The controller will then process known errors accordingly.

Currently that amounts to degrading/deleting misbehaving terminators and incrementing relevant service counters.


controller/network/network.go
```diff
@@ -463,6 +474,11 @@ func (network *Network) CreateCircuit(srcR *Router, clientId *identity.TokenId,
 		}
 		logger.Debugf("cleaned up [%d] abandoned routers for circuit", cleanupCount)
 
+		terminatorLocalAddress := peerData[uint32(ctrl_pb.ContentType_TerminatorLocalAddressHeader)]
+		path.TerminatorLocalAddr = string(terminatorLocalAddress)
+
+		delete(peerData, uint32(ctrl_pb.ContentType_TerminatorLocalAddressHeader))
+
```
plorenz, line 481: I forgot you were adding this. Very nice!

plorenz, line 481: I forgot you were adding this. Very nice!

controller/network/path.go
```diff
@@ -26,10 +26,11 @@ import (
...
-	Nodes     []*Router
-	Links     []*Link
-	IngressId string
-	EgressId  string
+	Nodes               []*Router
+	Links               []*Link
+	IngressId           string
+	EgressId            string
+	TerminatorLocalAddr string
 }
```
plorenz, line 35: It looks like the formatting is off a bit. Can you check if you're running gofmt on save?

plorenz, line 35: It looks like the formatting is off a bit. Can you check if you're running gofmt on save?

camotts: I am running format on save. What is the expected formatting here? It looks like I would have expected it, but maybe I'm expecting something different.

controller/network/path.go
```diff
@@ -44,6 +45,10 @@ func (self *Path) String() string {
 		out += fmt.Sprintf("->[l/%s]", self.Links[i].Id)
 		out += fmt.Sprintf("->[r/%s]", self.Nodes[i+1].Id)
 	}
```
plorenz, line 47: Can we include this in the last router field? I think otherwise it's going to be confusing. So rather than

`[r/lastRouterId] -> [r/localAddress]`  have something like

`[r/lastRouterId (localAddress)]`

plorenz, line 47: Can we include this in the last router field? I think otherwise it's going to be confusing. So rather than

`[r/lastRouterId] -> [r/localAddress]`  have something like

`[r/lastRouterId (localAddress)]`

controller/network/routesender.go
```diff
@@ -148,6 +126,61 @@ attendance:
 	return peerData, nil, nil
 }
```
plorenz, line 128: Above you've got counters for invalid terminator and misconfigured terminator, but I only see invalid being used here. I think you could mark invalid when you're deleting and misconfigured when you're setting to failed

plorenz, line 128: Above you've got counters for invalid terminator and misconfigured terminator, but I only see invalid being used here. I think you could mark invalid when you're deleting and misconfigured when you're setting to failed

controller/network/routesender.go
```diff
@@ -148,6 +126,61 @@ attendance:
...
 
+func (self *routeSender) handleRouteSend(attempt uint32, path *Path, strategy xt.Strategy, status *RouteStatus, terminator xt.Terminator, logger *pfxlog.Builder) (peerData xt.PeerData, cleanups map[string]struct{}, err error) {
+	if status.ErrorCode != nil {
```
plorenz, line 130: statusError != nil should only happend if success = false, so the code should reflect that in some way

plorenz, line 130: statusError != nil should only happend if success = false, so the code should reflect that in some way

router/handler_ctrl/route.go
```diff
@@ -156,13 +162,27 @@ func (rh *routeHandler) connectEgress(msg *channel.Message, attempt int, ch chan
...
+								errCode = ctrl_msg.ErrorTypeConnectionRefused
+							case syscall.ETIMEDOUT:
+								errCode = ctrl_msg.ErrorTypeDialTimedOut
+							}
+						}
+						headerError = &errCode
+					}
+					rh.fail(msg, attempt, route, errors.Wrapf(err, "error creating route for [c/%s]", route.CircuitId), headerError, log)
 				}
 			} else {
```
plorenz, line 181: Should return misconfigured terminator if the binding is invalid or the dialer can't be created

plorenz, line 181: Should return misconfigured terminator if the binding is invalid or the dialer can't be created

pb/ctrl_pb/impl.go
```diff
@@ -18,6 +18,14 @@ package ctrl_pb
 
 import "github.com/openziti/fabric/controller/xt"
```
plorenz, line 20: I'm thinking it's only used in the router, so this should live somewhere else. ctrl_pb is only for code related to ctrl channel communications data. It should probably live in xgress/ since it's a way for xgress components to communicate back to the dialer.

Since it's wrapping an error, it should also implement `Unwrap`, see https://pkg.go.dev/errors

plorenz, line 20: I'm thinking it's only used in the router, so this should live somewhere else. ctrl_pb is only for code related to ctrl channel communications data. It should probably live in xgress/ since it's a way for xgress components to communicate back to the dialer.

Since it's wrapping an error, it should also implement `Unwrap`, see https://pkg.go.dev/errors

controller/network/path.go
plorenz: I think maybe it's just the github view messing things up. If expand the section above the if block below, it shows the code indented 8, but that block shows indented 4. I'm guessing there's some difference in the GH formatter for expanded sections vs diff sections.

Also, not sure why all my comments got double posted :(

pb/ctrl_pb/impl.go
camotts: Moved to `xgress/messages.go`.  Added Unwrap to the error as well

router/handler_ctrl/route.go
camotts: Added MisconfiguredTerminator header when the dialer could not be created.
Also added another error wrapper to extend error checking above.

controller/network/routesender.go
camotts: I can change that. I must have misunderstood our last conversation with the counters. I thought we landed on just using invalid at the time, but I left both counters in thinking that would be fine. Going to change to use both as above

controller/network/path.go
camotts: Updated to append to last router rather than another addition

controller/network/path.go
```diff
@@ -41,13 +41,17 @@ func (self *Path) String() string {
 		return "{malformed}"
 	}
```
plorenz, line 42: It's a bit complicated b/c there may be zero links if we're initiating and terminating the connection on a single router.

camotts: I didn't think of that. Should have probably done it like this in the first place. Added a check if there are any links to conditionally print and added end of slice check to print within the loop.

controller/network/routesender.go
plorenz: Could add something which complains if the two aren't in agreement

camotts: Ooh I like that idea. Added a log message then process continues, unless you would also like it to return an actual error as well.

router/handler_ctrl/route.go
```diff
@@ -156,13 +161,30 @@ func (rh *routeHandler) connectEgress(msg *channel.Message, attempt int, ch chan
...
-				rh.fail(msg, attempt, route, errors.Wrapf(err, "unable to create dialer for [c/%s]", route.CircuitId), log)
+				var errCode byte = ctrl_msg.ErrorTypeMisconfiguredTerminator
+				rh.fail(msg, attempt, route, errors.Wrapf(err, "unable to create dialer for [c/%s]", route.CircuitId), &errCode, log)
 			}
 		} else {
-			rh.fail(msg, attempt, route, errors.Wrapf(err, "error creating route for [c/%s]", route.CircuitId), log)
+			rh.fail(msg, attempt, route, errors.Wrapf(err, "error creating route for [c/%s]", route.CircuitId), nil, log)
```
plorenz, line 187: This should also be misconfigured terminator

plorenz (approved):

router/handler_ctrl/route.go
camotts: Added. I don't exactly understand this one though. This would happen when a factory is needed that doesn't exist? I'm probably just thinking about it too much, but I don't see why that would be a terminator misconfiguration

plorenz: This would happen if the user created with a terminator with binding 'edge_tansplort' instead of 'edge_transport'. There would be no xgress Factory registered for that binding and you'd get that first error. But they could fix it by fixing the terminator binding.

PR #336, opened 2022-03-18
Replacing link 'type' with 'protocol' and 'costTags'

tburtchell:
Replaces the ambiguous 'type' property on a link listener with:

- **protocol:** Determined by the listener bind address.  tls/udp/etc
- **costTags:** Future placeholder for applying a base cost to links created via this listener

**Note** this removes the type-based multiple link feature added in 0.25.0.  This will be replaced soon by link groups and bind addresses.


controller/handler_ctrl/accept.go
```diff
@@ -94,18 +94,18 @@ func (self *CtrlAccepter) Bind(binding channel.Binding) error {
...
 					for _, listener := range listeners.Listeners {
-						log.WithField("address", listener.GetAddress()).WithField("type", listener.GetType()).Debug("router listener")
-						r.AddLinkListener(listener.GetAddress(), listener.GetType())
+						log.WithField("address", listener.GetAddress()).WithField("protocol", listener.GetProtocol()).WithField("costTags", listener.GetCostTags()).Debug("router listener")
+						r.AddLinkListener(listener.GetAddress(), listener.GetProtocol(), listener.GetCostTags())
 					}
 				}
 			} else if listenerValue, found := ch.Underlay().Headers()[channel.HelloRouterAdvertisementsHeader]; found {
 				log.Debug("router reported listener using advertisement header")
 				addr := string(listenerValue)
```
plorenz, line 103: Since we use exclusively tls links for now, should the default be tls?

controller/network/router.go
```diff
@@ -74,10 +74,11 @@ func (entity *Router) toBolt() boltz.Entity {
 	}
 }
```
plorenz, line 76: linkCostTags should probably be a `[]string`

pb/ctrl_pb/ctrl.proto
```diff
@@ -100,8 +100,9 @@ message Dial {
   string linkId = 1;
   string address = 2;
   string routerId = 3;
-  string linkType = 4;
+  string linkProtocol = 4;
   string routerVersion = 5;
```
plorenz, line 104: `repeated string linkCostTags = 6`

router/xlink_transport/config.go
```diff
@@ -55,11 +55,11 @@ func loadListenerConfig(data map[interface{}]interface{}) (*listenerConfig, erro
 		}
 	}
 
-	if value, found := data["type"]; found {
-		if typeString, ok := value.(string); ok {
-			config.linkType = typeString
+	if value, found := data["costTags"]; found {
```
plorenz, line 58: could either accept only `[]interface{}` or both `[]interface{}` and string (for single cost tag). In general we aren't that nice, so probably should do the first for consistency :)

controller/handler_ctrl/accept.go
tburtchell: Yeah.  I missed that.   I'll update.

dovholuknf (approved):

PR #341, opened 2022-03-29
Added heartbeat logging/metrics to control channel

camotts:


controller/handler_ctrl/bind.go
```diff
@@ -56,9 +64,21 @@ func (self *bindHandler) BindChannel(binding channel.Binding) error {
 	binding.AddPeekHandler(trace.NewChannelPeekHandler(self.network.GetAppId(), binding.GetChannel(), self.network.GetTraceController(), traceDispatchWrapper))
 	binding.AddPeekHandler(metrics2.NewCtrlChannelPeekHandler(self.router.Id, self.network.GetMetricsRegistry()))
```
plorenz, line 66: This should be the actual 0.25 version this will release with

controller/handler_ctrl/bind.go
```diff
@@ -99,3 +119,60 @@ func (self *ctrlChannelLatencyHandler) HandleLatency(latencyType latency.Type, e
...
+	}
+}
+
+func (self *heartbeatCallback) HeartbeatRx(int64) {}
+
+func (self *heartbeatCallback) HeartbeatRespTx(int64) {}
+
+func (self *heartbeatCallback) HeartbeatRespRx(ts int64) {
+	now := time.Now()
+	self.lastResponse = now.UnixMilli()
```
plorenz, line 144: This should be `now.UnixNano()` since we've already grabbed the time just above. I put in a fix for the link version when I saw this.

camotts: Moved to nano

controller/handler_ctrl/bind.go
camotts: I gotcha, I thought this was for when the heartbeat handler was added to make sure it worked. Using `0.25.5`

plorenz (approved):

plorenz (changes requested):
See below

controller/handler_ctrl/bind.go
```diff
@@ -99,3 +119,60 @@ func (self *ctrlChannelLatencyHandler) HandleLatency(latencyType latency.Type, e
...
+	}
+}
+
+func (self *heartbeatCallback) HeartbeatRx(int64) {}
+
+func (self *heartbeatCallback) HeartbeatRespTx(int64) {}
+
+func (self *heartbeatCallback) HeartbeatRespRx(ts int64) {
+	now := time.Now()
+	self.lastResponse = now.UnixNano()
```
plorenz, line 144: Sorry, I wasn't clear, I meant the time.Now().UnixNano() below should just be `now.UnixNano()`. The `now.UnixMillis()` should stay as it was

camotts: Oops, either would have made sense to me so didn't think much about it. Changed.

controller/handler_ctrl/bind.go
```diff
@@ -56,9 +64,21 @@ func (self *bindHandler) BindChannel(binding channel.Binding) error {
...
+	queueTimeHistogram := self.network.GetMetricsRegistry().Histogram("ctrl.queue_time:" + self.router.Id)
+	if doHeartbeat {
+		log.Info("router supports heartbeats")
+		cb := &heartbeatCallback{
+			latencyMetric:    roundTripHistogram,
+			queueTimeMetric:  queueTimeHistogram,
+			ch:               binding.GetChannel(),
+			latencySemaphore: concurrenz.NewSemaphore(2),
+		}
+		channel.ConfigureHeartbeat(binding, 10*time.Second, time.Second, cb)
```
plorenz, line 79: We need to have the same close handler to release the metrics as we do in the else block below. I need to check the link code if that's missing there as well. It likely is

plorenz: Yes, it's missing in the handler_link/bind.go as well. Would you mind adding it there as well, please?

camotts: Moved the close handler to be outside of the if/else block on the control channel part. Double checked the link side I think it's already done.

plorenz: :+1: Missed it in the link code, as I was looking at the bottom.

PR #345, opened 2022-03-30
Move foundation/config to fabric

plorenz:


config/env.go
```diff
@@ -0,0 +1,43 @@
+/*
```
plorenz, line 1: Exact copy of code from foundation

config/subconfig.go
```diff
@@ -0,0 +1,25 @@
+/*
```
plorenz, line 1: Exact copy of code from foundation

andrewpmartinez (approved):

PR #349, opened 2022-04-04
added circuit creation timespan

camotts:


plorenz:
Looks pretty good. I think you just need to add the value to the CircuitEvent type in fabric/events/circuits and update the PlainTextFabricCircuitEvent WriteTo method. The Json output should be handled automatically when you add the field.

plorenz (approved):

PR #350, opened 2022-04-04
Fix issue where peerdata could be set to nil if non-terminator router…

plorenz:
… return from route last


andrewpmartinez (approved):
Nice!

PR #356, opened 2022-04-13
threaded a timout through fabric

camotts:


plorenz:
I think you can drop the custom timeout type. So within an application we only care about the deadline. So you should be able to pass around `time.Time` as the deadline. Might make sense to encapsulate this in a Context, but that also might just obscure the thing we care about.

When passing the the timeout between apps, you'll need to translate the deadline into a timeout. So at the point that you send a message to another component, you figure out how much time is left till the deadline and send that as the timeout to other component.

Does that make sense? If not, we can talk it through

camotts:
> I think you can drop the custom timeout type. So within an application we only care about the deadline. So you should be able to pass around `time.Time` as the deadline. Might make sense to encapsulate this in a Context, but that also might just obscure the thing we care about.
> 
> When passing the the timeout between apps, you'll need to translate the deadline into a timeout. So at the point that you send a message to another component, you figure out how much time is left till the deadline and send that as the timeout to other component.
> 
> Does that make sense? If not, we can talk it through

Yep! All makes sense to me. Used a custom type here to be a bit more explicit, but if thats not necessary then all good!

controller/handler_ctrl/circuit_request.go
```diff
@@ -59,7 +60,7 @@ func (h *circuitRequestHandler) HandleReceive(msg *channel.Message, ch channel.C
 				}
 			}
 			log = log.WithField("serviceId", service)
-			if circuit, err := h.network.CreateCircuit(h.r, id, service, logcontext.NewContext()); err == nil {
+			if circuit, err := h.network.CreateCircuit(h.r, id, service, logcontext.NewContext(), time.Now().UTC().Add(network.DefaultTimeout)); err == nil {
```
plorenz, line 63: Does using UTC() buy you anything? Since none of the dealines are leaving the system, not sure where it would make a difference.

controller/network/network.go
```diff
@@ -753,7 +753,7 @@ func (network *Network) Run() {
 func (network *Network) handleLinkChanged(l *Link) {
 	log := logrus.WithField("linkId", l.Id)
 	log.Info("changed link")
```
plorenz, line 755: Should this take `deadline`?

plorenz: Ah, it's not part of CreateCircuit, nevermind :)

controller/network/options.go
```diff
@@ -31,6 +31,7 @@ const (
 	DefaultNetworkOptionsPendingLinkTimeout      = 10 * time.Second
 	DefaultNetworkOptionsSmartRerouteFraction    = 0.02
 	DefaultNetworkOptionsSmartRerouteCap         = 4
```
plorenz, line 33: We already have `DefaultNetworkOptionsRouteTimeout` above, I think we should probably use that

plorenz:
I had a few minor nits, but overall made sense, I think

controller/network/options.go
camotts: Sure! Wan't sure if we wanted more granularity here

plorenz:
One last thought. Can we check for a 0 deadline in the router and, if it's zero, either use the default timeout or don't set a timeout (based on whether the xgress impl dial has a configurable timeout)? 

So if we have a deadline, use the minimum of the deadline and the xgress impl timeout.
If we don't have a deadline, revert to the default xgress impl timeout behavior

plorenz (approved):

PR #358, opened 2022-04-19
Rework model for circuit inspections. Add inspect support back to mgm…

plorenz:
…t channel

### DIFF
--- a/controller/api_impl/inspections_api_model.go
+++ b/controller/api_impl/inspections_api_model.go
@@ -17,8 +17,10 @@
 package api_impl
 
 import (
+	"encoding/json"
 	"github.com/openziti/fabric/controller/network"
 	"github.com/openziti/fabric/rest_model"
+	"strings"
 )
 
 const EntityNameInspect = "inspections"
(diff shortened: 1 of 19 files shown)


controller/api_impl/inspections_api_model.go
```diff
@@ -29,10 +31,27 @@ func MapInspectResultToRestModel(inspectResult *network.InspectResult) *rest_mod
 		Success: &inspectResult.Success,
 	}
 	for _, val := range inspectResult.Results {
+		var emitVal interface{}
+		if strings.HasPrefix(val.Value, "{") {
+			mapVal := map[string]interface{}{}
```
ekoby, line 36: is `HasPrefix()` preferable to `if val.Value[0] == '{'`

router/xgress/link_receive_buffer.go
```diff
@@ -84,9 +85,9 @@ func (buffer *LinkReceiveBuffer) getLastBufferSizeSent() uint32 {
 	return atomic.LoadUint32(&buffer.lastBufferSizeSent)
 }
 
-func (buffer *LinkReceiveBuffer) Inspect() map[string]interface{} {
-	result := map[string]interface{}{}
-	result["size"] = buffer.Size()
-	result["lastSizeSent"] = buffer.getLastBufferSizeSent()
-	return result
+func (buffer *LinkReceiveBuffer) Inspect() inspect.XgressRecvBufferDetail {
```
ekoby, line 88: return `*inspect.XgressRecvBufferDetail` for consistency with `LinkSendBuffer.Inspect()`

router/xgress/xgress.go
```diff
@@ -580,16 +581,15 @@ func (self *Xgress) SendEmptyAck() {
...
+func (self *Xgress) InspectCircuit(detail *inspect.CircuitInspectDetail) {
 	timeSinceLastRxFromLink := time.Duration(info.NowInMilliseconds()-atomic.LoadInt64(&self.timeOfLastRxFromLink)) * time.Millisecond
-	result["timeSinceLastLinkRx"] = timeSinceLastRxFromLink.String()
-	result["sendBuffer"] = self.payloadBuffer.Inspect()
-	result["recvBuffer"] = self.linkRxBuffer.Inspect()
+	xgressDetail := &inspect.XgressDetail{
+		Address:             string(self.address),
+		Originator:          self.originator.String(),
+		TimeSinceLastLinkRx: timeSinceLastRxFromLink.String(),
+		SendBufferDetail:    *self.payloadBuffer.Inspect(),
```
ekoby, line 590: any reason to not have detail fields pointers?

ekoby (approved):

router/xgress/link_receive_buffer.go
plorenz: Fixed

router/xgress/xgress.go
plorenz: I liked guaranteeing they wouldn't be null, but they're always going to be non-nil just based on usage, so probably makes sense to just use pointers. Updated.

PR #360, opened 2022-04-26
Fix link race condition

plorenz:


router/xlink_transport/listener.go
```diff
@@ -224,7 +224,7 @@ func (self *listener) bindNonSplitChannel(binding channel.Binding, routerId, rou
 
 	if existingLink, applied := self.xlinkRegistery.LinkAccepted(xli); applied {
 		log.Info("link registered")
-	} else {
+	} else if existingLink != nil {
```
dovholuknf, line 227: does it make sense to keep an actual else indicating neither match for "unexpected situations'?

router/handler_ctrl/dial.go
```diff
@@ -103,7 +103,7 @@ func (self *dialHandler) handle(dial *ctrl_pb.Dial, _ channel.Channel) {
 				if err := self.sendLinkMessage(dial.LinkId); err != nil {
 					log.WithError(err).Error("error sending link message ")
 				}
-			} else {
+			} else if existingLink != nil {
```
dovholuknf, line 106: same "catchall 'else' comment here"

controller/network/network.go
```diff
@@ -330,6 +330,11 @@ func (network *Network) LinkConnected(id string, connected bool) error {
 
 	if l, found := network.linkController.get(id); found {
 		if connected {
+			if state := l.CurrentState(); state != nil && state.Mode != Pending {
+				log.Infof("link is %v not pending, cannot mark connected", state.Mode)
```
dovholuknf, line 334: maybe "link **state** is %v..."? making it a bit more clear in the log?

controller/network/network.go
```diff
@@ -330,6 +330,11 @@ func (network *Network) LinkConnected(id string, connected bool) error {
...
+				return nil
```
dovholuknf, line 335: why return nil - why not return the error (and then not log)

router/xlink_transport/listener.go
plorenz: The other condition (not applied, no existing link) is logged in LinkAccepted

controller/network/network.go
plorenz: fixed

controller/network/network.go
plorenz: Fixed. Removed all logging from this method, since we were logging in calling methods.

dovholuknf (approved):

PR #408, opened 2022-05-27
uses generic version of cmap

andrewpmartinez:


dovholuknf (approved):
This feels "So Good™"

PR #410, opened 2022-05-31
Adding Metrics API on '/metrics'

tburtchell:


plorenz:
Unless you prefer not to, can I request that you either squash when merging or squash before merging just to keep the history a bit tidier? It's a lot of commits :)

controller/api_impl/metrics-api.go
```diff
@@ -0,0 +1,175 @@
...
+	return metricsApiHandler, nil
+}
+
+func NewMetricsApiHandler(n *network.Network, options map[interface{}]interface{}) (*MetricsApiHandler, error) {
+	metricsApi := &MetricsApiHandler{
+		options:    options,
+		network:    n,
+		inspectMgr: network.NewInspectionsManager(n),
+	}
+
```
plorenz, line 77: I don't think we do inline certs anywhere else. Should this just be a path to cert file?

controller/api_impl/metrics-api.go
```diff
@@ -0,0 +1,175 @@
...
+func (metricsApi *MetricsApiHandler) IsHandler(r *http.Request) bool {
+	return strings.HasPrefix(r.URL.Path, metricsApi.RootPath())
+}
+
+func (metricsApi *MetricsApiHandler) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
+	metricsApi.handler.ServeHTTP(writer, request)
+}
+
+func (metricsApi *MetricsApiHandler) newHandler() http.Handler {
```
plorenz, line 144: For consistency, no blank lines at the start of functions, please

controller/api_impl/metrics-api.go
tburtchell: certainly could be.  I'll look at making that change.

plorenz (approved):

PR #412, opened 2022-06-06
move terminator address to seperate field

camotts:
https://github.com/openziti/ziti/issues/565


camotts:
> I don't see a new field on the event. Am I missing something?

I might have misunderstood the intent here. I was under the impression we wanted to change the log only. I can go add it to the event as well.

plorenz:
> > I don't see a new field on the event. Am I missing something?
> 
> I might have misunderstood the intent here. I was under the impression we wanted to change the log only. I can go add it to the event as well.

I think the main goal is to have the path and the address available in the event, just as separate fields.

camotts:
> > > I don't see a new field on the event. Am I missing something?
> > 
> > 
> > I might have misunderstood the intent here. I was under the impression we wanted to change the log only. I can go add it to the event as well.
> 
> I think the main goal is to have the path and the address available in the event, just as separate fields.

Definitely misunderstood then! Will add that

plorenz (approved):

plorenz:
Looks good :+1:

PR #428, opened 2022-06-29
Remove management channel

plorenz:


rest_util/capool.go
```diff
@@ -0,0 +1,88 @@
+/*
+	Copyright NetFoundry, Inc.
```
andrewpmartinez, line 2: I'm not sure what the differences are as you this was  straight copy-paste.

Can you highlight what you needed to change? Or if that is too much speculation on only having 1 version of this code?

plorenz: This one I left alone. It's not actually used by the fabric bits, but retrieving and verifying controller certs seems like a generically useful thing, so I thought maybe it should live in fabric. 
I did  just notice that it's hitting an edge endpoint, so maybe I need to take out the verify and leave only the retrieval function? Or maybe take out both, since they're not actually used?

In util.go, I stripped out some bits that were edge specific and left the bits that were used by the edge.

clients.go is pretty much all new and fabric specific. I tried using the interfaces you'd defined (Authenticator) for edge, but they didn't make sense for fabric since the fabric needs to be able to use pure cert based, but also piggy back on edge auth when available. I thought it would be straightforward to implement a TransportFactory in terms of the apis you have in the edge.

dovholuknf: It's `NetFoundry Inc.` no comma...

plorenz: @dovholuknf we should run a separate cleanup script, as the `,` is fairly widespread

andrewpmartinez: Second the secondary cleanup. It will blow this PR up.

andrewpmartinez: Mmmm I say we just move forward with this and consolidate over time. I don't think it is worth the brain power to find the best solution. This is good enough and moves us in the right direction.

andrewpmartinez (approved):

PR #448, opened 2022-07-15
Refactor events subsystem. Add link events. Fixes #415

### LINKED ISSUE #415: Add link events
Events for 

* link dialed (with local ip/port remote ip/port) - sent to ctrl from router
* link accepted (with local ip/port remote ip/port) -sent to ctrl from router
* link created
* link faulted

plorenz:
Previously we often had two sets of events, one in the network
package, and then another in the events package. We now only
have one, in the events package. Events now have a public API,
 with minimal deps. There's an implementation package which has
the dependencies. This allows the network package to generate
events, without have a circular dependency on the events
implementation package.


michaelquigley (approved):
:+1:

PR #461, opened 2022-07-29
Add flags to create/update commands. allows some lightweight logic sp…

plorenz:
…ecialization (eg priveleged vs non-priveleged updates)


andrewpmartinez (approved):

andrewpmartinez:
I assume the flags will have their values defined at a later date/some other place.

plorenz:
> I assume the flags will have their values defined at a later date/some other place.

The idea is that each type could have it's own flags. So CaManager could use a 'verified' flag to indicate that we were updating the Ca to indicate it was verified, and that would be the only time the verified flag could be updated.

PR #504, opened 2022-09-19
Readonly lock

camotts:
Added a readonly lock to prevent commands being run by mismatched controllers.

Up'd the go version to 1.19 to take advantage of new typed atomics.

### DIFF
--- a/controller/raft/mesh/mesh.go
+++ b/controller/raft/mesh/mesh.go
@@ -17,6 +17,11 @@
 package mesh
 
 import (
+	"net"
+	"sync"
+	"sync/atomic"
+	"time"
+
 	"github.com/hashicorp/raft"
 	"github.com/michaelquigley/pfxlog"
 	"github.com/openziti/channel"
@@ -25,14 +30,12 @@ import (
 	"github.com/openziti/transport/v2"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
-	"net"
-	"sync"
-	"time"
 )
 
 const (
-	PeerIdHeader   = 10
-	PeerAddrHeader = 11
+	PeerIdHeader      = 10
+	PeerAddrHeader    = 11
+	PeerVersionHeader = 12
 
 	RaftConnectType = 2048
 	RaftDataType    = 2049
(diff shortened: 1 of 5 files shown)


controller/raft/mesh/mesh.go
```diff
@@ -109,9 +116,12 @@ type Mesh interface {
 	// GetOrConnectPeer returns a peer for the given address. If a peer has already been established,
 	// it will be returned, otherwise a new connection will be established
 	GetOrConnectPeer(address string, timeout time.Duration) (*Peer, error)
+	IsReadOnly() bool
 }
 
-func New(id *identity.TokenId, raftId raft.ServerID, raftAddr raft.ServerAddress, bindHandler channel.BindHandler) Mesh {
+func New(id *identity.TokenId, version string, raftId raft.ServerID, raftAddr raft.ServerAddress, bindHandler channel.BindHandler) Mesh {
```
plorenz, line 122: I think you can just embed a non-pointer atomic.Bool in the impl. It won't need to be initialized b/c the default value will be false.

controller/raft/mesh/mesh.go
```diff
@@ -230,6 +246,12 @@ func (self *impl) AddPeer(peer *Peer) {
 	self.lock.Lock()
 	defer self.lock.Unlock()
 	self.Peers[peer.Address] = peer
```
plorenz, line 248: Comment is misleading, not checking for higher, just equality

controller/raft/raft.go
```diff
@@ -119,6 +122,10 @@ func (self *Controller) Dispatch(cmd command.Command) error {
 		}
 	}
 
+	if self.Mesh.IsReadOnly() {
```
plorenz, line 125: Can you add some text about why it's read-only? (either 'upgrade in progress' or 'different versions detected in cluster'). Leaning towards different versions, since that's most explicit.

Don't need Errorf because you're not injecting any params. Can just use error.New
Also general convention is lowercase for errors.

controller/raft/mesh/mesh.go
camotts: changed to non-pointer

controller/raft/mesh/mesh.go
camotts: Yea that makes sense. As implemented it would also readonly if an older version joined

plorenz (approved):

PR #512, opened 2022-09-27
Support multiple control channels. Fixes #469

### LINKED ISSUE #469: Initial support for multiple control channels in routers

plorenz:


events/dispatcher_usage.go
```diff
@@ -116,7 +116,7 @@ func (self *usageEventAdapter) AcceptMetricsMsg(message *metrics_pb.MetricsMessa
 				for usageType, usage := range bucket.Values {
 					evt := &event.UsageEvent{
 						Namespace:        event.UsageEventsNs,
-						Version:          event.UsageEventsVersion,
+						Version:          2,
```
michaelquigley, line 119: magic number?

router/config.go
```diff
@@ -101,7 +103,7 @@ type Config struct {
 		}
 	}
 	Ctrl struct {
-		Endpoint              *UpdatableAddress
+		Endpoints             []*UpdatableAddress
```
michaelquigley, line 106: definitely would love to see the updated controllers get written to a state file. 👍

michaelquigley (approved):
👍

events/dispatcher_usage.go
plorenz: Usually have a constant for this, but we're supporting two versions at the same time, so one gets the constant and the other is hard coded to the old version. At point we'll remove support for the old version.

PR #529, opened 2022-10-24
Always use a new router instance for connected routers so we can ensure

plorenz:
disconnect handling applies to the right instance


michaelquigley (approved):
Seems reasonable...

PR #536, opened 2022-11-11
Various HA related fixes

### LINKED ISSUE #525: Update metrics message propagation from router to controller for HA
Send metrics messages to all controllers. All controllers need them so they can extract latency.
Set flag on message to indicate it should be propagated to external events. Once a controller has accepted the message, unset the flag, so only one controller propagates the metrics.
This will ensure that 1 and only 1 controller propagates the metrics. Small window where server can down after receiving message but before propagating, but it's a very small window.

plorenz:
* Router cert handling fix
* Always start with empty bolt db, so raft journal can be applied cleanly
* Allow looking up raft member ids, so you only need to provide addr
* Report metrics to all controllers. Fixes #525


michaelquigley (approved):
Looks sensible to me. 👍

andrewpmartinez (approved):

PR #537, opened 2022-11-15
Router controller updates

camotts:
Updated router config to export and endpoints state file which holds all known controller ctrl channels. This gets updated as new controllers are added


camotts:
This will send out any known ctrl channel address to a router when it connects (if raft is not enabled it will send just the one controller's address).

Whenever a new controller is brought online the "main" controller's routers will get immediate updates. Once the raft consensus goes through it will trigger updates on each other controller's routers.

controller/settings.go
```diff
@@ -47,3 +47,35 @@ func (o OnConnectSettingsHandler) RouterConnected(r *network.Router) {
 		}).Info("no on connect settings to send")
 	}
 }
+
+type OnConnectCtrlAddressesUpdateHandler struct {
```
plorenz, line 51: I see how the callback works, but I think it would be cleared if you just had the network in here and had the GetAddresses as a method on the handler

controller/settings.go
```diff
@@ -47,3 +47,35 @@ func (o OnConnectSettingsHandler) RouterConnected(r *network.Router) {
...
+	callback func() []string
+}
+
+func (o *OnConnectCtrlAddressesUpdateHandler) RouterDisconnected(r *network.Router) {
+	//do nothing, satisfy interface
+}
+
```
plorenz, line 58: This should be a pointer receiver for consistency

controller/settings.go
```diff
@@ -47,3 +47,35 @@ func (o OnConnectSettingsHandler) RouterConnected(r *network.Router) {
...
+func (o OnConnectCtrlAddressesUpdateHandler) RouterConnected(r *network.Router) {
+	pfxlog.Logger().Info("Router connected... syncing crtl addresses")
+	data := o.callback()
+	pfxlog.Logger().Info(data)
+	updMsg := &ctrl_pb.UpdateCtrlAddresses{
+		Addresses: data,
+	}
+
```
plorenz, line 66: If you add a GetContentType for `UpdateCtrlAddresses` in ctrl_pb/impl.go, you can do 

```
protobufs.MarshalTyped(updMsg).Send(r.Control)
```

controller/settings.go
```diff
@@ -47,3 +47,35 @@ func (o OnConnectSettingsHandler) RouterConnected(r *network.Router) {
...
+type OnConnectCtrlAddressesUpdateHandler struct {
+	callback func() []string
+}
+
+func (o *OnConnectCtrlAddressesUpdateHandler) RouterDisconnected(r *network.Router) {
+	//do nothing, satisfy interface
+}
+
+func (o OnConnectCtrlAddressesUpdateHandler) RouterConnected(r *network.Router) {
```
plorenz, line 59: I'd probably do 
```
log := pfxlog.Logger().WithField("routerId", r.Id
```
at the top here and that way you don't have to do it at every point of use

controller/controller.go
```diff
@@ -391,3 +402,27 @@ func (c *Controller) Identity() identity.Identity {
 func (c *Controller) GetEventDispatcher() event.Dispatcher {
 	return c.eventDispatcher
 }
+
+func (c *Controller) routerDispatchCallback() error {
```
plorenz, line 406: Only called from raftController so nil check shouldn't be necessary, right?

controller/controller.go
```diff
@@ -391,3 +402,27 @@ func (c *Controller) Identity() identity.Identity {
...
+	if c.raftController != nil {
+		data := c.raftController.CtrlAddresses()
+		updMsg := &ctrl_pb.UpdateCtrlAddresses{
+			Addresses: data,
+			IsLeader:  c.raftController.IsLeader(),
+			Index:     c.raftController.GetRaft().LastIndex(),
+		}
```
plorenz, line 413: See comment below about updMsg and protobufs.MarshallTyped

controller/controller.go
```diff
@@ -391,3 +402,27 @@ func (c *Controller) Identity() identity.Identity {
 func (c *Controller) GetEventDispatcher() event.Dispatcher {
 	return c.eventDispatcher
 }
+
```
plorenz, line 405: I would have this take the raft.Configuration directly and work with that. That way you don't have to worry about the change not having propagated yet.

plorenz: Ah, I see you're storing it and then retrieving it, so you don't have to worry about syncing state, but since the only place is used is in this method, I still think it's cleaner to pass it through.

controller/raft/member.go
```diff
@@ -1,17 +1,17 @@
 /*
```
plorenz, line 1: Can you revert this change? It's the only change to the file and it's inconsistent with the copyright blocks in other files

controller/raft/mesh/mesh.go
```diff
@@ -369,3 +370,14 @@ func (self *impl) checkState() {
 func (self *impl) IsReadOnly() bool {
 	return self.readonly.Load()
 }
+
```
plorenz, line 373: Method is unused, please remove

router/config.go
```diff
@@ -314,6 +317,29 @@ func (c *UpdatableAddress) Store(address transport.Address) {
 	c.wrapped.Store(address)
 }
```
plorenz, line 319: Marshall/UnmarshallYaml should have pointer receivers for consistency with the other methods

router/router.go
```diff
@@ -524,6 +543,75 @@ func (self *Router) RegisterXWebHandlerFactory(x xweb.ApiHandlerFactory) error {
...
+	}
+
+	b, err := os.ReadFile(endpointsFile)
+	if err != nil {
+		return nil
+	}
+
+	if err := yaml.Unmarshal(b, &self.ctrlEndpoints); err != nil {
+		return err
+	}
```
plorenz, line 570: What kind of mismatches?

router/router.go
```diff
@@ -524,6 +543,75 @@ func (self *Router) RegisterXWebHandlerFactory(x xweb.ApiHandlerFactory) error {
...
+			pfxlog.Logger().Info("Adding new endpoint")
+			save = true
+			parsed, err := transport.ParseAddress(ep)
+			if err != nil {
+				return err
+			}
+			self.ctrlEndpoints.Set(ep, NewUpdatableAddress(parsed))
+		}
+	}
+
```
plorenz, line 592: Should probably always do this check, just in case we added an endpoint and removed one at the same time, so the lengths are the same, but the contents are different

router/router.go
```diff
@@ -524,6 +543,75 @@ func (self *Router) RegisterXWebHandlerFactory(x xweb.ApiHandlerFactory) error {
...
+	if len(self.ctrlEndpoints.ConcurrentMap) != len(endpoints) {
+		for knownep := range self.ctrlEndpoints.Items() {
+			if _, ok := newEps[knownep]; !ok {
```
plorenz, line 595: Log which endpoint is being removed

router/router.go
```diff
@@ -524,6 +543,75 @@ func (self *Router) RegisterXWebHandlerFactory(x xweb.ApiHandlerFactory) error {
...
+				pfxlog.Logger().Info("Removing old endpoint")
+				save = true
```
plorenz, line 597: Also make sure to close control channel when removing

router/config.go
```diff
@@ -458,6 +484,11 @@ func LoadConfig(path string) (*Config, error) {
 					return nil, errors.Wrap(err, "invalid value for ctrl.defaultRequestTimeout")
 				}
 			}
```
plorenz, line 486: For now, if dataDir isn't defined, can we define it as the parent directly of the config file?

router/handler_ctrl/bind.go
```diff
@@ -37,9 +38,10 @@ type bindHandler struct {
 	ctrlAddressChanger CtrlAddressChanger
 	xgDialerPool       goroutines.Pool
 	linkDialerPool     goroutines.Pool
+	ctrlAddressUpdater CtrlAddressUpdater
 }
```
plorenz, line 43: I'd like to consolidate the address changer and updated functionality, but let's do that in a follow up PR

plorenz:
Overall looks reasonable, lots of nitpicks :)

router/env/ctrl.go
```diff
@@ -173,3 +189,14 @@ func (self *networkControllers) Close() error {
 	})
 	return errors.ToError()
 }
+
+func (self *networkControllers) CloseAndRemoveByAddress(address string) error {
+	ctrlId, ch := self.GetCtrlChannelByAddress(address)
+	if ch != nil {
+		if err := ch.Close(); err != nil {
+			return err
```
plorenz, line 197: I think probably always want to delete, even if there is an error

router/handler_ctrl/updateCtrlAddresses.go
```diff
@@ -0,0 +1,50 @@
...
+	if err := proto.Unmarshal(msg.Body, upd); err != nil {
+		log.WithError(err).Error("error unmarshalling")
+		return
+	}
+
+	if upd.IsLeader || handler.currentVersion == 0 || handler.currentVersion < upd.Index {
+		if err := handler.callback.UpdateCtrlEndpoints(upd.Addresses); err != nil {
+			log.WithError(err).Error("Unable to update ctrl addresses")
+		}
+		handler.currentVersion = upd.Index
```
plorenz, line 38: Maybe log the addresses here as well, unless they're already getting logged elsewhere?

camotts: So it's getting logged in the `UpdateCtrlEndpoints`. Can add it here as a bulk message, in the aforementioned its logs them individually when they are being added/removed. Saying that out loud does tell me that already known endpoints are hidden so perhaps also log here as well actually?

router/env/ctrl.go
camotts: Can do!

router/router.go
```diff
@@ -482,6 +515,60 @@ func (self *Router) connectToController(addr *UpdatableAddress, localBinding str
...
 }
 
+func (self *Router) connectToControllerWithBackoff(addr *UpdatableAddress, localBinding string, maxTimeout *time.Duration) error {
+	self.controllersToConnect.mtx.Lock()
+	log := pfxlog.Logger()
+	defer func() {
+		self.controllersToConnect.mtx.Unlock()
+	}()
+	if _, exists := self.controllersToConnect.controllers[addr]; exists {
+		log.WithField("Controller Addr", addr).Info("Already atempting to connect")
```
plorenz, line 525: atempting -> attempting

router/router.go
```diff
@@ -482,6 +515,60 @@ func (self *Router) connectToController(addr *UpdatableAddress, localBinding str
 	return nil
 }
```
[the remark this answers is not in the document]
camotts: That is something that I didn't really consider. Looking through that retry package I'm not seeing anything that intends to allow us to cancel the retries. It looks like we could modify the `MaxElapsedTime` to be a small number so the next iteration fails, or we could create our own exponential backoff that is cancellable (and potentially create an issue to see if its something that would benefit the original library)

router/handler_ctrl/updateCtrlAddresses.go
plorenz: I think it's probably fine. As long as the change is logged somewhere not sure the entire set needs to be logged each time

router/router.go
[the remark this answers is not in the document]
camotts: Meant to mention this earlier. Added a delete from the connecting map when a ctrl channel is deleted. Also added that error

router/router.go
```diff
@@ -535,6 +535,9 @@ func (self *Router) connectToControllerWithBackoff(addr *UpdatableAddress, local
 	}
 
 	operation := func() error {
```
plorenz, line 537: Still needs to be protected by lock. Might be simpler to just use a cmap.ConcurrentMap. Depends on if you're using the lock just for concurrent access or if you're also protecting a critical section.

plorenz (approved):

PR #557, opened 2022-12-16
Add minimum cost delta for smart rerouting. Fixes #552

### LINKED ISSUE #552: Add minimum cost delta for smart routing
Add configurable min cost delta for smart routing, so we only switch the route if the cost improvement is > x (where x is configurable)

Add configuration under network/smart with a default value of 15.

```
network:
  smart:
    minCostDelta: 20 # default 15
```

plorenz:


andrewpmartinez (approved):

controller/network/smart_test.go
```diff
@@ -0,0 +1,118 @@
...
+	link1.SetStaticCost(20)
+
+	candidates = network.getRerouteCandidates()
+	assert.Equal(t, 0, len(candidates))
+
+	// updated cost would be 52, hit threshold, should change
+	link1.SetStaticCost(21)
+
+	candidates = network.getRerouteCandidates()
+	assert.Equal(t, 1, len(candidates))
```
ekoby, line 117: check the right path in candidates list?

ekoby (approved):

controller/network/smart_test.go
plorenz: fixed

PR #570, opened 2022-12-28
Support auto-migrate and agent initiated migration. Fixes #569

### LINKED ISSUE #569: Support automatic migration and agent based migration

plorenz:


controller/controller.go
```diff
@@ -391,3 +378,61 @@ func (c *Controller) Identity() identity.Identity {
...
+
+func (c *Controller) InitializeRaftFromBoltDb(sourceDbPath string) error {
+	log := pfxlog.Logger()
+
+	log.Info("waiting for raft cluster to settle before syncing raft to database")
+	start := time.Now()
+	for !c.raftController.IsLeader() {
+		time.Sleep(time.Second)
+		if time.Since(start) > time.Second*30 {
+			log.Panic("cannot sync raft to database, as current node is not the leader")
```
ekoby, line 407: should this case return an error? Does it get into an infinite loop if it fails to get leadership?

plorenz: Switched it to only require cluster to have a leader, not that local node is leader. It will panic after 30s if the cluster doesn't elect a leader.

ekoby (approved):

PR #574, opened 2022-12-30
Fix some handler removal. Only propagate crud terminator events on le…

### LINKED ISSUE #573: Ensure specific events aren't duplicated in raft cluster
* fabric.circuits - unique to each controller
* fabric.links - some overlap, probably unavoidable. Will be addressed if/when we move more link mgmt to routers 
* fabric.routers - about ctrl <-> router connection, so will likely be similar but not identical across controllers
* fabric.terminators - create/update/delete events should only happen from leader by default
* metrics and fabric.usage - first controller to receive propagates 
* services - different per controller. will need to be aggregated externally like usage
* edge.entityCounts - only generate on leader by default
* edge.sessions/edge.apiSessions - TBD, likely going away

plorenz:
…ader by default. Fixes #573

### DIFF
--- a/controller/command/command.go
+++ b/controller/command/command.go
@@ -43,11 +43,16 @@ type Validatable interface {
 // system is the leader, apply it locally
 type Dispatcher interface {
 	Dispatch(command Command) error
+	IsLeaderOrLeaderless() bool
 }
 
 // LocalDispatcher should be used when running a non-clustered system
 type LocalDispatcher struct {
 	EncodeDecodeCommands bool
+}
+
+func (self *LocalDispatcher) IsLeaderOrLeaderless() bool {
+	return true
 }
 
 func (self *LocalDispatcher) Dispatch(command Command) error {
--- a/events/dispatcher_terminator.go
+++ b/events/dispatcher_terminator.go
@@ -26,6 +26,7 @@ import (
 	"github.com/pkg/errors"
 	"go.etcd.io/bbolt"
 	"reflect"
+	"strings"
 	"time"
 )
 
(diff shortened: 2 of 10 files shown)


controller/command/command.go
```diff
@@ -43,13 +43,18 @@ type Validatable interface {
 // system is the leader, apply it locally
 type Dispatcher interface {
 	Dispatch(command Command) error
+	IsLeaderOrLeaderless() bool
```
andrewpmartinez, line 46: This naming could imply that being a leader is different than being leaderless. 

Just curious what the paths are to becoming leaderless. Is it just that an existing leader can go offline?

events/dispatcher_terminator.go
```diff
@@ -45,14 +54,29 @@ func (self *Dispatcher) AcceptTerminatorEvent(event *event.TerminatorEvent) {
...
 
-	self.AddTerminatorEventHandler(handler)
+	propagateAlways := false
+	if val, found := options["propagateAlways"]; found {
+		if b, ok := val.(bool); ok {
+			propagateAlways = b
+		} else if s, ok := val.(string); ok {
+			propagateAlways = strings.EqualFold(s, "true")
+		} else {
+			return errors.New("invalid value for propagateAlways, must be boolean or string")
```
andrewpmartinez, line 71: String?

andrewpmartinez (approved):

controller/command/command.go
plorenz: Yes, I just wanted to account for a state where a leader had gone offline or we were in the middle of an election for anyreason. I figured occasional extra events were better than occasional missed events.

PR #580, opened 2023-01-05
Added raft leave and raft list members

camotts:


controller/agent.go
```diff
@@ -172,7 +175,33 @@ func (self *Controller) agentOpRaftRemove(m *channel.Message, ch channel.Channel
 	//	return err
 	//}
 	// _, err := c.WriteString("success\n")
-	handler_common.SendOpResult(m, ch, "raft.remove", "no yet implemented", false)
+
```
plorenz, line 178: Should be able to specify addr or id, shouldn't require addr.

camotts: Updated, originally done that way to have join/leave have the same input structure

camotts: Saying that makes zero sense here, that was mostly for the cli side which previously (before my change) required sending the address so I made it required here before too

controller/network/network.go
```diff
@@ -95,7 +98,7 @@ type Network struct {
 	serviceMisconfiguredTerminatorCounter     metrics.IntervalCounter
 }
```
plorenz, line 100: The network already has the raft controller. It's available as the command.Dispatcher instance. I think in this case it would make sense for the RaftRouter to get that and if it's not nil, try to cast it to `*raft.Controller`. Does that make sense?

camotts: Makes sense to me after that explanation!

plorenz (approved):

PR #600, opened 2023-02-09
Miscellaneous raft related updates

plorenz:
Add initial cluster events.
Simplify ctrl <-> ctrl connection.
Fix version mismatch detection

### DIFF
--- a/controller/config.go
+++ b/controller/config.go
@@ -292,16 +292,12 @@ func LoadConfig(path string) (*Config, error) {
 					if val, found := submap["advertiseAddress"]; found {
 						if advertiseAddr, ok := val.(string); ok {
 							if advertiseAddr != "" {
-								if addr, err := transport.ParseAddress(advertiseAddr); err == nil {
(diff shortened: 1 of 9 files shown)


controller/config.go
```diff
@@ -298,7 +298,11 @@ func LoadConfig(path string) (*Config, error) {
 									return nil, fmt.Errorf("error loading advertiseAddress for [ctrl/options] (%v)", err)
 								}
 								if controllerConfig.Raft != nil {
-									controllerConfig.Raft.AdvertiseAddress = advertiseAddr
+									addr, err := transport.ParseAddress(advertiseAddr)
+									if err != nil {
+										return nil, errors.Wrap(err, "unable to parse [ctrl/options/advertiseAddress]")
```
dovholuknf, line 303: can youprint what the value was that couldn't parse too? seems like a friendly thing to do imo

dovholuknf (approved):

controller/config.go
plorenz: Fixed

PR #605, opened 2023-02-14
Delay link.dropped_msgs metric init until channel bind to prevent met…

### LINKED ISSUE #604: Don't create link dropped msg metric until channel bind time
Otherwise, if channel creation fails, we don't have a hook to clean things up, leading to a metrics leak.

plorenz:
…ric leak. Fixes #604


dovholuknf (approved):
Seems fine to me

PR #620, opened 2023-03-01
Configurable Usage (and Metrics) Reporting Interval

michaelquigley:
Default to the current behavior, but allow both routers and the controller to be be configured to utilize a different usage and metrics reporting frequency.


router/metrics/peekhandler.go
```diff
@@ -159,7 +160,7 @@ func NewXgressPeekHandler(registry metrics.UsageRegistry) xgress.PeekHandler {
 		egressTxMsgSizeHistogram:  egressTxMsgSizeHistogram,
 		egressRxMsgSizeHistogram:  egressRxMsgSizeHistogram,
```
plorenz, line 162: Can you add this to the config as well, please? Otherwise it's just a constant in a new place :)

michaelquigley: i'm not specifically sure what you're pointing at?

router/router.go
```diff
@@ -480,6 +485,7 @@ func (self *Router) startControlPlane() error {
 	}
 
 	self.metricsReporter = fabricMetrics.NewControllersReporter(self.ctrls)
+	logrus.Warnf("report interval = %v", self.config.Metrics.ReportInterval)
```
michaelquigley, line 488: can also get rid of this.

plorenz (approved):

PR #652, opened 2023-03-24
Regenerate rest client with swagger 0.30.4

plorenz:


dovholuknf (approved):
is this thing versioned at all? is that relevant ?

PR #653, opened 2023-03-24
router disable

### LINKED ISSUE #651: Add router enable/disable mechanism
Disabled routers should be disconnected and no new connections should be allowed

plorenz:
- Regenerate rest client with swagger 0.30.4
- Add router disable flag. Fixes #651


pb/cmd_pb/cmd.proto
```diff
@@ -97,7 +97,8 @@ message Router {
   bytes fingerprint = 3;
   uint32 cost = 4;
   bool noTraversal = 5;
-  map<string, TagValue> tags = 6;
```
andrewpmartinez, line 100: Re-ordering fields breaks backwards compat, no?

andrewpmartinez (approved):

pb/cmd_pb/cmd.proto
plorenz: Correct. However, not going to worry about it until 1.0 release, as there are no current install using raft.

PR #705, opened 2023-05-11
Add config support for link groups. Doesn't use group config yet.

plorenz:


router/xlink_transport/config.go
```diff
@@ -118,6 +127,16 @@ func loadDialerConfig(data map[interface{}]interface{}) (*dialerConfig, error) {
 		}
 	}
 
+	if value, found := data["groups"]; found {
```
dovholuknf, line 130: `groups` should be a constant imo

dovholuknf (approved):
lgtm. are additional attributes a reason to bump the version? (i don't know how that's accomplished in this repo)

plorenz:
> lgtm. are additional attributes a reason to bump the version? (i don't know how that's accomplished in this repo)

At least this portion of the work is backwards compatible, so I wasn't going to bump the version yet.

PR #716, opened 2023-05-19
Refactor bootstrap. Cleanup some error messages. Set ctrl check to initial fail state

plorenz:


controller/controller.go
```diff
@@ -437,7 +437,7 @@ func (c *Controller) TryInitializeRaftFromBoltDb() error {
 	path := fmt.Sprintf("%v", val)
 	if _, err := os.Stat(path); err != nil {
 		if os.IsNotExist(err) {
-			return errors.Wrapf(err, "source db not found at [%v]", path)
+			return errors.Wrapf(err, "source db not found at [%v], either remove 'db' config setting or fix path ", path)
```
andrewpmartinez, line 440: As a heads up, `%v` can hide logging issues. If `path` is changed from a `string` to a `*string` for some reason, this code will not show up in linting and instead will happily print memory addresses.

When possible, I would suggest not using `%v` when there are specific flags for printing the proper type (%s, %d, etc).

andrewpmartinez (approved):

PR #726, opened 2023-05-24
Notify controller of forward faults on links. Fixes #724. If reroute …

### LINKED ISSUE #724: Controller should be notified of forwarding faults on links

### LINKED ISSUE #725: If reroute fails, circuit should be torn down

plorenz:
…fails, circuit should be torn down. Fixes #725


controller/network/fault.go
```diff
@@ -17,33 +17,37 @@
...
-		} else {
 			// unroute non-existent circuit
 			if err := sendUnroute(ffr.R, circuitId, true); err == nil {
-				logrus.Infof("sent unroute for [s/%s] to [r/%s] in response to forwarding fault", circuitId, ffr.R.Id)
+				log.Info("sent unroute for circuit to router in response to forwarding fault")
 			} else {
-				logrus.Errorf("error sending unroute for [s/%s] to [r/%s] in response to forwarding fault (%v)", circuitId, ffr.R.Id, err)
+				log.WithError(err).Error("error sending unroute for circuit to router in response to forwarding fault")
 			}
 		}
```
andrewpmartinez, line 52: `if/else` changed to `if/else if`. If the 3rd unhandled case (found ==false and UnknownOwner== true) occurs there is no logic to handle it.

I leave it up to your knowledge. I don't know if that case is necessary to handle.

controller/network/network.go
```diff
@@ -119,7 +119,7 @@ func NewNetwork(config Config) (*Network, error) {
 		routerChanged:         make(chan *Router, 16),
 		linkController:        newLinkController(config.GetOptions()),
 		linkChanged:           make(chan *Link, 16),
-		forwardingFaults:      make(chan *ForwardingFaultReport, 16),
+		forwardingFaults:      make(chan struct{}, 1),
```
andrewpmartinez, line 122: Curious to understand why a buffer of 16 faults and now 1?

router/forwarder/faulter.go
```diff
@@ -88,19 +88,32 @@ func (self *Faulter) run() {
...
+					if ch == nil {
+						log.Error("no control channel for controller")
+						continue
+					}
+
+					fault := &ctrl_pb.Fault{Subject: ctrl_pb.FaultSubject_ForwardFault, Id: circuitIds}
+					if err := protobufs.MarshalTyped(fault).Send(ch); err == nil {
+						log.WithField("circuitCount", len(workload)).Warn("reported forwarding faults")
+					} else {
+						log.WithError(err).Error("error sending fault report")
```
andrewpmartinez, line 104: What happens in this condition (we have a specific controller to forward a fault to and can't send it to them)?

I see the fault is dropped. I assume if the destination controller comes back, I assume the circuit w/ the invalid forward is eventually abandoned by the client, the client has made a new circuit (most likely through a new controller), and the fault-containing circuit is eventually cleaned up.

andrewpmartinez (approved):

andrewpmartinez (approved):

controller/network/fault.go
plorenz: If the circuit is unknown but we don't definitely own it, then there's nothing we can do.

controller/network/network.go
plorenz: It's now just a signal, indicating that cleanup should be done. It was running the fault stuff in a goroutine already, so now it just does that first, before signalling.

router/forwarder/faulter.go
plorenz: If the circuit continues to fault, we'll have another chance to send. If it doesn't, then presumably the issue was resolved in some other way.

PR #736, opened 2023-06-02
Link check should be initially passing if min links is zero. Fixes #735

### LINKED ISSUE #735: router link health check should only be passing initially if min links is zero

plorenz:


router/config.go
```diff
@@ -664,22 +666,30 @@ func LoadConfig(path string) (*Config, error) {
...
 						if val, ok := value.(int); ok {
 							cfg.HealthChecks.LinkCheck.MinLinks = val
 						} else {
 							return nil, errors.Wrapf(err, "invalid value [%v] for healthChecks.linkCheck.minLinks", value)
 						}
 					}
 
+					if value, found := checkMap["initialDelay"]; found {
+						if val, err := time.ParseDuration(fmt.Sprintf("%v", value)); err == nil {
+							cfg.HealthChecks.LinkCheck.InitialDelay = val
```
dovholuknf, line 687: accompanying doc/examples are updated elsewhere i take it, right?

router/router.go
```diff
@@ -586,8 +586,8 @@ func (self *Router) initializeHealthChecks() (gosundheit.Health, error) {
 	err = h.RegisterCheck(&linkHealthCheck{router: self, minLinks: checkConfig.LinkCheck.MinLinks},
 		gosundheit.ExecutionPeriod(checkConfig.LinkCheck.Interval),
 		gosundheit.ExecutionTimeout(5*time.Second),
-		gosundheit.InitiallyPassing(true),
-		gosundheit.InitialDelay(1*time.Second),
+		gosundheit.InitiallyPassing(checkConfig.LinkCheck.MinLinks == 0),
```
dovholuknf, line 589: a thought... feels like logic that should be in a separate function that's eventually testable/assert-able?

dovholuknf (approved):
looks ok to me...

router/router.go
plorenz: Yes. I would wait to refactor it until we add the test, other wise it'll likely need to be refactored twice :)

router/config.go
plorenz: Yes, examples are in ziti

PR #742, opened 2023-06-06
amqp fix

### LINKED ISSUE #732: ampq event shipper should not crash the controller
The ampq shipper is a great addition to allow event streaming to external systems, but it really needs the following two improvements to be viable in a production setting. 

1. Connection failure upon startup should not prevent the controller from starting, it should log an error and then continue
2. Failure to ship events should result in the data being buffered up to a configurable size while the connection retries, but if the buffer fills up the event shipping should log errors that it can't buffer anymore. 

Key idea - Metrics and event shipping should never crash the controller, and needs to be able to handle slow shipping, connection failures, auth errors, timeouts, etc.

camotts:
updated amqp writer to connect in the background and reconnect on failures. 

fixes #732

### DIFF
--- a/events/amqp_logger.go
+++ b/events/amqp_logger.go
@@ -19,9 +19,12 @@ package events
 import (
 	"context"
 	"fmt"
+	"time"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/pkg/errors"
 	amqp "github.com/rabbitmq/amqp091-go"
+	"github.com/sirupsen/logrus"
 )
 
 type AMQPEventLoggerFactory struct{}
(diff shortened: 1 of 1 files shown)


events/amqp_logger.go
```diff
@@ -31,40 +34,134 @@ func (AMQPEventLoggerFactory) NewEventHandler(config map[interface{}]interface{}
...
+				}
+			}
+		}
+	}()
+
+	return ret
+}
+
+func (wc *amqpWriteCloser) sendMessage(message []byte) {
+	select {
```
plorenz, line 94: Unlikely, but you could still blow out the stack if you recurse enough times. 

```
for {
   trySend
   if success return
   if !closed {
      log
      return
   }
   reconnect
}
```

camotts: Good catch, better to be safe here

plorenz (approved):

PR #757, opened 2023-07-13
adds server certs to raft mesh peer connections

andrewpmartinez:


plorenz (approved):
Looks good. I'm not clear on what the certs will be used for, but I'm sure it'll be come clear in a future PR :)

PR #758, opened 2023-07-14
Move link calculation to routers.

plorenz:


controller/idgen/id.go
```diff
@@ -56,9 +56,10 @@ func (self *shortIdGenerator) NextAlphaNumericPrefixedId() (string, error) {
...
 
-func NewUUIDString() (string, error) {
+func NewUUIDString() string {
 	id := uuid.New()
 	v := &big.Int{}
 	v.SetBytes(id[:])
-	return basex.EncodeInt(v)
+	result, _ := basex.EncodeInt(v)
+	return result
 }
```
andrewpmartinez, line 65: Doesn't `uuid` support `uuid.NewString()` or does this do something different.

plorenz: It uses a denser encoding (fewer characters) than the default uuid encoding

router/xlink_transport/dialer.go
```diff
@@ -34,6 +34,30 @@ type dialer struct {
...
+	adoptedBinding     string
+}
+
+func (self *dialer) GetHealthyBackoffConfig() xlink.BackoffConfig {
+	return self.config.healthyBackoffConfig
+}
+
+func (self *dialer) GetUnhealthyBackoffConfig() xlink.BackoffConfig {
+	return self.config.unhealthyBackoffConfig
+}
```
andrewpmartinez, line 46: What are healthy vs unhealthy dial-backoffs?

I assume healthy is for dials that are known to be available to some other part of the deployment and unhealthy are when no components can reach them, but that is a guess.

andrewpmartinez (approved):

router/xlink_transport/dialer.go
plorenz: healthy is when the target router is connected to the controller, unhealthy is when it's not. If it's not connected to the controller, it's likely down, but it may not be, so it will still try to connect, just using a different (likely more aggressive) set of backoff parameters

PR #784, opened 2023-09-01
Cluster connections need ALPN protocol when dialing. Fixes #783

### LINKED ISSUE #783: Raft cluster connections not updated for ALPN
Cluster connections are getting denied because they're not specifying an ALPN protocol

plorenz:


dovholuknf (approved):
lgtm

PR #795, opened 2023-09-15
Allow specifying snapshot name with templating. Fixes #794 Fixes #792

### LINKED ISSUE #792: include raft index in DB snapshot filename

### LINKED ISSUE #794: Add output file to ziti agent controller snapshot-db command
The current command ziti agent controller snapshot-db generates a file names based on a timestamp, but that timestamp is based on when is successfully completes the snapshot & so is unpredictable.

Example:  you issue the command on 20230104-062503, but the output file can be different depending on how long it takes to take the snapshot, so the output file can be 20230104-062503+n(number of minutes it can take), so you end up with 20230104-062805(took 3 minutes)

It would be nice to add a `--output` so you can specify the name of the file AND/OR have the output of the command return the file name instead of just "success"

plorenz:


controller/db/db.go
```diff
@@ -42,3 +44,12 @@ func Open(path string) (boltz.Db, error) {
 
 	return db, nil
 }
+
+func LoadCurrentMetadat(tx *bbolt.Tx, name string) uint64 {
```
andrewpmartinez, line 48: Is the metadata just the index?

Also, `Metadat` to `Metadata` ?

andrewpmartinez (approved):

controller/db/db.go
plorenz: Fixed. I'll switch it to GetCurrentMetadata when there are other consumers. There is other stuff stored under metadata, but that's currently the only value that's retrieved with that method.

PR #797, opened 2023-09-15
Make link heartbeats configurable. Add unresponsive close timeout. Fixes #796

### LINKED ISSUE #796: Make link heartbeats configurable, including an unresponive close timeout

plorenz:


router/handler_link/bind.go
```diff
@@ -166,11 +164,19 @@ func (self *heartbeatCallback) HeartbeatRespRx(ts int64) {
 func (self *heartbeatCallback) CheckHeartBeat() {
 	log := pfxlog.Logger().WithField("channelId", self.ch.Label())
 	now := time.Now().UnixMilli()
-	if self.firstSent != 0 && (now-self.firstSent > 30000) && (now-self.lastResponse > 30000) {
+	if delta := now - self.lastResponse; delta > 30000 {
```
ekoby, line 167: is 30000 not the configurable value?

router/handler_link/bind.go
```diff
@@ -166,11 +164,19 @@ func (self *heartbeatCallback) HeartbeatRespRx(ts int64) {
...
 		log.Warn("heartbeat not received in time, link may be unhealthy")
 		self.latencyMetric.Clear()
 		self.latencyMetric.Update(88888888888)
```
ekoby, line 170: maybe `math.MaxUInt64`?

ekoby (approved):

router/handler_link/bind.go
plorenz: I'll add this in a second pass, but it requires more plumbing b/c the default heartbeat options type doesn't have a space for it, so I need to create a custom object and thread it through.

router/handler_link/bind.go
plorenz: It just needs to be sufficiently large and a marker value, so also easy to recognize when someone is looking at it. Before this logic was added we never updated the latency if no response was received, so we could be reporting low values. I'll add a couple more eights, which will bump it up to 2 hours.

